PR #14, opened 2021-04-02
[caa edits] Add fatbox to non-square packagings

kellnerd:
Tested with [Edit #78562367](https://musicbrainz.org/edit/78562367), front covers should usually have the same aspect ratio as back covers for regular jewel cases and can possibly also include spines.
I also checked the [docs](https://musicbrainz.org/doc/Release/Packaging) for more likely non-square packaging types while I was at it, but I found none. Only the _Snap Case_ could be another possible candidate - but I have never seen one and there is no example image, only a description:

> A digipak-like case held together with a "snapping" plastic closure.

### DIFF
--- a/mb_supercharged_caa_edits.user.js
+++ b/mb_supercharged_caa_edits.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Supercharged Cover Art Edits
-// @version      2021.4.1.2
+// @version      2021.4.2
 // @description  Supercharges reviewing cover art edits. Displays release information on CAA edits. Enables image comparisons on removed and added images.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
@@ -62,6 +62,7 @@ const NONSQUARE_PACKAGING_TYPES = [
     6, // Keep case
     8, // Cassette
     9, // Book
+    10, // Fatbox
     17, // Digibook
 ];
 


kellnerd:
By the way, while I was testing on my branch I intuitively wanted to install the test version using the install button. Of course this did not work because these are currently using absolute links.

If you would replace them by relative links this would simplify testing on branches other than the `main` branch. In order to convert the raw links to relative links there is an undocumented query parameter which you can use: https://github.com/kellnerd/musicbrainz-bookmarklets/commit/e227ce1fff9e273dc6b8864558cb31e0af460e7a

ROpdebee:
> If you would replace them by relative links this would simplify testing on branches other than the `main` branch.

Great trick, thanks!

> Only the Snap Case could be another possible candidate

I found some images of snap cases on Discogs: [1](https://www.discogs.com/kd-lang-Sexuality/release/116832), [2](https://www.discogs.com/Prodigy-Firestarter/release/102379). Seems indeed to be non-square, so I'll add those as well. Gatefold covers _might_ be another one, but I'm not entirely sure. I seem to recall seeing one recent case where it wasn't square.

kellnerd:
For gatefolds you would probably have to distinguish between CDs and Vinyl, gatefold packagings for CDs are usually non-square while those for Vinyl are square.

ROpdebee:
True. Similarly, backs and trays of normal jewel cases aren't square either. But handling all of those edge cases would massively increase the complexity of the check, the ugliness of the code, or both 😞. Currently, it's probably not worth putting in the effort, IMO, but it could be done in the future. As long as it manages to find a considerable number of dubious covers, I'm happy 🙂

PR #15, opened 2021-04-02
Run a spell checker and fix typos

kellnerd:
Just found a typo in the README and decided to run the spell checker of my IDE over the code as I still had it open. In case you are using VS Code I can also share the workspace dictionary with some words I marked as exceptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Blinds editor and voter details before your votes are cast.
 
 ## MB: Bulk copy-paste work codes
 
-Quickly copy-paste work idenitifiers (ISWC, agency work codes) from CISAC's ISWCNet into a MusicBrainz work.
+Quickly copy-paste work identifiers (ISWC, agency work codes) from CISAC's ISWCNet into a MusicBrainz work.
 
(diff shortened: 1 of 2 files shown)


ROpdebee:
Thank you for notifying me that I can't spell 😄 
I'm also seriously wondering why I misspelled "agency" as "agncy" thrice, because it really does seem like it was intentional, but it doesn't shadow any other variables or anything so.. I'm really clueless!

kellnerd:
Maybe the variable name auto-completion of your IDE is to blame? Once mistyped and then subsequently you have always chosen the suggested name because intelli-sense can't be wrong 😇

> it doesn't shadow any other variables or anything

Sorry, in order to safe you some work I should have mentioned that I did not just blindly accept the spell checker's suggestions or use search and replace, I used a refactoring action (or three to be more precise) to rename all occurrences of the variables.

ROpdebee:
> in order to safe you some work I should have mentioned that I did not just [...]

No no no, don't worry, I mentioned shadowing as a potential reason why I named it so weirdly, but that was evidently not the case. More than likely, I went with "agncy" because the value isn't actually the "agency", it's the agency name, and just naming it "agency" would not represent that properly, and "agencyName" was too long for my liking. That's some very stupid reasoning, but knowing myself, it's probably true. 😂

PR #34, opened 2021-09-21
Fix import of metadata files

kellnerd:
- Correct the file extension
- Import the module using a file URL for compatibility with Windows, otherwise drive letters are recognized as an invalid protocol

I could not resist and had to try out the new features of mb_upload_to_caa_from_url, so I tried to setup the build environment on my computer ❤️

### DIFF
--- a/rollup/plugin-userscript.ts
+++ b/rollup/plugin-userscript.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import { pathToFileURL } from 'url';
 
 import type { Plugin } from 'rollup';
 import type { PackageJson } from 'type-fest';
@@ -119,7 +120,8 @@ function _userscript(options: Readonly<_UserscriptOptionsWithDefaults>): Plugin 
      * @return     {Promise<UserscriptMetadata>}  The userscript's metadata.
      */
     async function loadMetadata(): Promise<AllUserscriptMetadata> {
-        const metadataFile = path.resolve('./src', options.userscriptName, 'meta.js');
+        // use file URLs for compatibility with Windows, otherwise drive letters are recognized as an invalid protocol
+        const metadataFile = pathToFileURL(path.resolve('./src', options.userscriptName, 'meta.ts')).href;
         const specificMetadata: UserscriptMetadata = (await import(metadataFile)).default;
         return insertDefaultMetadata(specificMetadata);
     }


ROpdebee:
(closed and reopened to trigger CI)

ROpdebee:
Thanks! Hope you like the new features :) I'm aiming to get them released before the end of the week, the main thing that's left to do is to add buttons to import from URLs attached to the release.

kellnerd:
Of course I like the new features, now the script is basically doing all the things I was doing by hand all the time :) I thought it would take you longer to publish these new features, since my impression was that you wanted to port all of your userscripts before merging this branch, which still seems to be a fair bit of work.

So I checked out this branch, tried to install the dependencies and run the build script myself, which turned out to be not as straight forward as I initially thought:
- NPM refused to install the `github:ROpdebee/nativejsx` dependency with `npm ci` and I had to manually install it afterwards.
- The build script failed with this error image for all userscripts although the files are present in the repo:
  https://github.com/ROpdebee/mb-userscripts/blob/afb202d53c94a973c3c54db8ef431ecb5b2984e8/rollup.ts#L77-L85 My first thought was that this was also problem with the handling of absolute paths on Windows (that was before I found the problem I have adressed in this PR), but it turned out that [Node.js 14 LTS does not support `Promise.any()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any#browser_compatibility) and the catch-clause had absorbed this error.
- After upgrading to Node.js 16 the build script was finally running far enough for me to discover the metadata bug and with the new NPM version `npm ci` was also able to resolve the `github:ROpdebee/nativejsx` dependency correctly.
- Finally I had to look for places in the code where a module gets imported with an absolute path as the error message of the ESM loader was not very helpful and only showed the line numbers in `ts-node` where the error occurred...
- The last hurdle was the file extension of the `meta.ts` modules, not sure why this did not cause any trouble for you. Did you still have an old untracked `meta.js` file from before the TS refactoring in your working directory?

After building and opening the compiled userscripts the huge amount of code which was generated by the typescript conversion and the polyfills (mostly JSX?) kind of surprised me. In case you want to make this build process appealing to me as well, the amount of foreign code should definitely be less massive, right now I only buy the advantages of typing 😜

ROpdebee:
Yeah I realised that it's better to port the scripts gradually whenever the situation calls for it. In this case, the port was (IMO) well justified.

I'd eventually like to set up a system so that CI compiles the script on every commit so that it's not necessary to compile it all the time, but that's some future work :)

As for the foreign code, most of it are babel helpers for classes and async/await etc. JSX should be fairly lightweight, as most of it is translated into fairly simple function calls at compile time. Same goes for TypeScript, the build simply removes all type information rather than having TS compile the code, the only thing I know of where it actually changes something is for enums.

Ultimately though, even the original code is barely recognisable after transpilation, because all classes are transpiled by babel for browser support. I've opted to support all browsers that MB supports, so it has to transpile/polyfill a lot of features :/ Because it's so unrecognisable, I've decided to change the build to minify all of the code. That leads to a 50% reduction in file size, which I think is worth it considering that the previously-unminified transpiled code wasn't readable either.

This build system should definitely be improved further, but currently I'm mainly concerned with getting this script out there for everyone to enjoy. It'll be released shortly 🎉

PR #39, opened 2021-09-23
Fix checks for image duplicates

kellnerd:
Dealing with proper URLs instead of URL strings is not always easier, https://github.com/ROpdebee/mb-userscripts/commit/4ae985d439410515ee2fd4c80625131d86d5094c broke the comparison of image URLs and made the use of a `Set` kind of pointless:
Two `URL` objects for the same image URL are not considered identical, so I changed this to store and compare the `URL.href` attributes (of type string) instead.

In addition to that I have two other minor improvements:
- Remove the .js extension from type definition imports and adapt tsconfig.json to make the type acquisition of VS Code happy. There are still four regular imports with a "wrong" .js extension left, but removing the extension causes the build script to fail (for whatever reason).
- Simplify the calculation of the longest metadata field's length (just nitpicking, but more efficient and easier understandable IMO).


tsconfig.json
```diff
@@ -11,6 +11,10 @@
         "noImplicitReturns": true,
         "pretty": true,
         "baseUrl": "./",
+        "typeRoots": [
+            "node_modules/@types",
+            "types",
+        ],
         "paths": {
             "*": [
                 "types/*",
```
kellnerd, line 20: Not sure whether `typeRoots` makes `paths` redundant as you seem to use it only for type definitions and always have proper relative or absolute paths for regular module imports. I have not removed it since it probably does the trick in your IDE, but VS Code needs the `typeRoots` property and acts dumb without it.

rollup/plugin-userscript.ts
```diff
@@ -31,8 +31,7 @@ const DEFAULT_OPTIONS = {
 };
 
 function _userscript(options: Readonly<_UserscriptOptionsWithDefaults>): Plugin {
-    const longestMetadataField = [...options.metadataOrder]
-        .sort((a, b) => b.length - a.length)[0];
+    const longestMetadataFieldLength = Math.max(...options.metadataOrder.map((field) => field.length));
```
kellnerd, line 34: I have to say that I really enjoyed reading (all) your typescript code, you have a really elegant and clean coding style 👍
But this line has been giving me a headache ever since I came across it the first time, so I took the opportunity to improve it while I was already editing this module 😇

ROpdebee:
Looks good to me, but could you merge in the latest main and check the `wasMaximised` calculation? I think I just fixed that separately when working on the edit notes.

The whole URL string -> URL conversion was mainly driven by concerns that the script would be fed URLs with parameters or hashes, e.g. `https://discogs.com/release/12345?foo=bar` and that such URLs might break some of the regexp checks. Using `url.pathname` avoids that, and it felt better to do the conversion at the earliest possible point rather than temporarily converting it all across the code (but then again, now I have to do `url.href` a lot, but at least I get early validation of invalid URLs, I guess)

The funny thing is that I knew full-well that `URL` doesn't properly do equality checks (and IMHO it's really stupid that it doesn't), I encountered that in 04c765368e06539ad77c17a796222cc8b917340c :)

I've also go no clue what's with those JS extensions, probably something to do with babel, typescript, node, and the mixing of CJS and ES modules, and trying until something worked. Then changing something else, which then made all of the trial and error useless. Ugh.

src/mb_enhanced_cover_art_uploads/index.tsx
```diff
@@ -179,14 +180,14 @@ class ImageImporter {
...
+        if (this.#doneImages.has(url.href)) {
             this.#banner.set(`${originalFilename} has already been added`);
             this.#clearInput(url.href);
             return;
         }
-        this.#doneImages.add(url);
+        this.#doneImages.add(url.href);
 
-        let result;
+        let result: FetchResult;
```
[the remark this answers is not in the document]
ROpdebee: Do you happen to know of a setting that forces me to write types for such declarations? I get an eerily uncomfortable feeling seeing bare `let`s where the type of the variable isn't either explicitly specified or fairly easy to see from the initialiser.

tsconfig.json
ROpdebee: No clue. I started doing this with no TS experience, and I've never properly RTFD. The config is sub-optimal anyway, it's leading to implicitly included types which IMO should be implicitly included (like `__CoverArt` in https://github.com/ROpdebee/mb-userscripts/blob/main/src/lib/MB/__MB__.d.ts, which aliases another type in the src). I'll tinker with it a bit once everything slows down :)

rollup/plugin-userscript.ts
ROpdebee: 100% agree. Absolute horrific. I have nothing to say to defend myself. 😄

kellnerd:
> Looks good to me, but could you merge in the latest main and check the wasMaximised calculation? I think I just fixed that separately when working on the edit notes.

Yes, I noticed this right before I created the pull request. But since you have changed that line in exactly the same way there are no merge conflicts with this PR. I just merged locally and it auto-merged fine indeed. I will push the merge commit now but I guess it just creates unnecessary clutter in the git history, so I can withdraw it again before you merge.

> I've also go no clue what's with those JS extensions, probably something to do with babel, typescript, node, and the mixing of CJS and ES modules, and trying until something worked. Then changing something else, which then made all of the trial and error useless. Ugh.

Probably somehow related to the fact that these are (in)directly imported by the build script itself while the extensionless imports are only touched by rollup!? Never change a running system 😂

tsconfig.json
kellnerd: I discovered https://www.typescriptlang.org/tsconfig/#types while I was looking for the docs of `typeRoots`. Probably you need to overwrite that option to disable these automatic imports

src/mb_enhanced_cover_art_uploads/index.tsx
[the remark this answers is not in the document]
kellnerd: No idea, `"noImplicitAny": true` did not seem to have any effect on these, `tsc` did not even complain about them after relaunching the IDE...

ROpdebee:
>Probably somehow related to the fact that these are (in)directly imported by the build script itself while the extensionless imports are only touched by rollup!? Anyway, never change a running system 😂

https://github.com/facebook/hhvm/blob/39415630e84538fb553cd325a85ce837bcf6cb70/src/runtime/ext/ext_imagesprite.cpp#L649

I think I need some tests as well 😛

src/mb_enhanced_cover_art_uploads/index.tsx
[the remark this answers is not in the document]
kellnerd: To be honest, I also have very little experince with TypeScript, but after having read through the possible `tsconfig.json` options, I also checked which options ESLint has to offer (including those of `@typescript-eslint`) and found the rule `'init-declarations': ['warn', 'always']` at least.
If we would enable this, it would at least trigger a warning for each declaration without initialisation, which makes it straight-forward to find the candidates for variables without inferred types or type annotations. Then we would have to mark them as exceptions:
```ts
// eslint-disable-next-line init-declarations
let result: FetchResult;
```
What do you think? Is it worth the ugly ESLint comments?

Thank you also very much for the invitation to become a collaborator and for the trust that I will not mess this repo up (I already added labels to this PR to test one of my new priviledges and failed at the first attempt due to the fat finger syndrome 😂).
If you want, I can push these changes (I see you are enjoying your experiments with the CI setup in a separate branch). I also have already prepared a commit that enables ESLint also for .jsx and .tsx files (after I wondered why my changes had no effect on the outcome of the linter).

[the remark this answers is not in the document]
ROpdebee: I always like a good hack like that, but I despise the ESLint ignore comment format (it's way too long, `flake8`'s `# noqa` ftw), so let's not 🙂

> I also have already prepared a commit that enables ESLint also for .jsx and .tsx files (after I wondered why my changes had no effect on the outcome of the linter).

Yes please. That explains why it wasn't complaining about [this unused import](https://github.com/ROpdebee/mb-userscripts/commit/32c5f0a0b44056f9b0cd8b46cc718bcb60aad7a0#diff-625513237b12a2d47cfd00590a1a8a2cefcb9cdbf2080df59c3e6ef614ab6a29L3) previously. I thought it had something to do with the `lib/*` ignore pattern (which is meant to ignore `./lib/`, not `./src/lib`), but you found the real cause

> Thank you also very much for the invitation to become a collaborator and for the trust that I will not mess this repo up

You're of course very welcome to contribute or help maintain the repo, but please don't feel pressured to do so. I added you as a collaborator so that if anything urgent needs to be addressed and I'm not around to do it for whatever reason, that at least one other person has write access, and you're the obvious choice. Don't interpret this as a nudge of "hey fix these bugs and implement these features" 😛 In regards to messing up the repo, I think we should still stick to PRs, and I'll be switching over to using feature branches and PRs myself too. I'll probably prohibit direct pushes to `main` as well to further enforce that. It's a better practice IMO, and I intend to eventually switch over to a Continuous Delivery system where a release is made on each PR merge, 1 PR = 1 feature/bug fix. It's pretty much what I'm mostly doing already anyway, but manually, so automating that seems like the right choice.

On that topic
> I see you are enjoying your experiments with the CI setup in a separate branch

Working on CI is so much fun, said no one ever. But it's going pretty well so far. The goal is to have a system that builds and commits to the `dist/` directory on each version increment in a `meta.ts` automatically, so that's a first step into the CD direction. Eventually it should increment versions itself though, but I'm not doing that until there are reliable tests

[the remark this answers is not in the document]
kellnerd: Ok, so I will push the minor commit https://github.com/ROpdebee/mb-userscripts/commit/ade55719e0322fbd57638eb456b2f4484fa80a41 directly to `main` and we drop https://github.com/ROpdebee/mb-userscripts/commit/b96db12c8f22fe2b738a671e1a5b9767361fd2c8 , alright?

tsconfig.json
ROpdebee: It's a bit weird though, since `typeRoots` apparently only matters for `/// <reference types=".." />` directives (per [here](https://github.com/Microsoft/TypeScript/issues/12222#issuecomment-260417733). To my knowledge, we're never using those directives.

Anyway, if it works for you, I'm fine with it :) I'll be changing the paths anyway and might be splitting up some of the `tsconfig` files, since I've got a feeling something's not working properly.

kellnerd: I am just reviewing your changes from #68 and apparently the `typeRoots` property is indeed not necessary (now). Removing it from `tsconfig.json` and restarting VS Code does no longer cause problems with relative `import type` statements. Something must have changed in the mean time, either your recent refactoring or me updating the dependencies has solved the problem. Anyway, I'm happy to remove it again once we are done with these huge PRs.

PR #44, opened 2021-09-23
Run ESLint for .tsx files & Use it to spot variables without inferred types

kellnerd:
- Run ESLint for .jsx and .tsx files too (was missing in `package.json` and `.eslintrc.cjs`).
- Also includes fixes for most of these new linter warnings (except any type in catch clauses).
- Enable warnings for variable declarations without initialisation. This makes it easier to ensure they have type annotations (to prevent them from being any type) before disabling the warning locally.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -46,7 +46,7 @@ module.exports = {
     },
     overrides: [
         {
-            files: ['*.ts'],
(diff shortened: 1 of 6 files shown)


ROpdebee:
Awesome, thanks!

I'll see what I can do about the remaining warnings

PR #45, opened 2021-09-23
Automatically distribute new userscript versions

ROpdebee:
Adapts CI to push the built userscripts if the version number changed since a previous commit.

Each update is given its own commit (no commit is created when the version number wasn't changed).

Two caveats:
1. This breaks if there's a force-push to `main`. However, since `main` will be protected from direct pushes, that bug should never occur.
2. The major downside is that these are pushed to `main`, so you might have to merge `main` into feature branches more often. On the other hand, those merges should always be fast-forwards/rebases, and shouldn't conflict with anything.


kellnerd (commented):
Looks sensible to me, at least the code parts (I know nothing about GitHub CI).

ROpdebee:
Hold on, I've protected `main` in such a way that direct pushes are not allowed, not sure whether that applies to GH Action pushes as well...

ROpdebee:
I'm closing this for now. It's getting way too late for me to fix this, and the approach is suboptimal anyway.

I think I've reached a possible solution wherein the job on the main branch creates a release branch and pushes the automated release commit to that new branch; and then a separate workflow run on the release branch does the checks required for the main branch, merges itself into the main branch after the checks complete, thereby bypassing the requirement that pushed commits have the check completed. However, it doesn't work in an automated way, because workflows can't run on commits created by other workflows.

Three possible solutions:
1. Use a custom token in the first workflow, that will allow the second workflow to run on the release commit, but keeps the spaghetti.
2. Don't require checks on pushes to main, that eliminates the need for the second workflow but makes it possible to mess up main
3. Don't distribute the compiled code in the source tree and use releases instead. I'd prefer this, but I'm a bit hesitant since I've already changed the script location earlier this week. OTOH, I'm sure I'll switch to releases eventually, and the longer I wait, the more script locations I'll have to change in the future.

ROpdebee:
Actually, working with releases will be a mess as well. Each time a script is updated, we'd need to copy the compiled files of all other unchanged scripts as well, and versions would need to be unique across all scripts. There are some alternatives but they go against the core idea of releases.

Option 4, which I'm preferring now, is to use a separate `dist` branch used only for the bot to commit on and people to fetch updates from, which is completely disassociated from `main`.

ROpdebee:
Implemented the separate `dist` branch idea and clarified the force-push problem. I've done a small test and it seems to work properly, it successfully pushed to `dist` (I've already reverted that change though).

Post-merge, the file in the `dist` folder in `main` should be recompiled with a version increment so that auto-updates can be used to transparently switch to the new location.

kellnerd (approved):
I had exactly the same thoughts on options 3 and 4 but did not have the time to write them down. The changes look good to me, also remember to update the URLs in the README when you do the version bump in `main/dist`.

PR #49, opened 2021-09-24
Use GM_info.namespace as homepage URL

kellnerd:
Fixes an issue with Tampermonkey where the edit note contains `undefined` instead of the hompage URL, see [the report in the forum](https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/20) and my reply.

The `namespace` property has the widest support among userscript engines, while `homepageURL` and `homepage` are only available in Violentmonkey and Tampermonkey respectively.


ROpdebee:
Thanks!

PR #61, opened 2021-09-25
Include `mb_collapse_work_codes` on paginated artist pages

### LINKED ISSUE #59: `mb_collapse_work_codes` doesn't run on second page of works
Reported by Tiske Tisja via DM:

> On page 1 of the works everything collaps like it should with this addon. However on the rest of the pages it does not.
Do you understand what I mean? Hard for me to explain in English
Example: See page 1 and page 2 of the works of dolly parton

My first guess is that it's missing some `@match` stanzas, but it might also be react-related (I can't remember how other pages are loaded).

ROpdebee:
I'm not a huge fan of this approach, since it will now match `artist/.../workPageThatShouldNotBeProcessed`, but it's the only way I can find to fix this without using `@include` instead.

Tested on FF/TM and FF/VM and confirmed to work on both. I'm assuming it should work on other combinations too.

This shouldn't be an issue for other `@match` patterns in this script as they all end in `*`.

Fixes #59

### DIFF
--- a/mb_collapse_work_attributes.user.js
+++ b/mb_collapse_work_attributes.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Collapse Work Attributes
-// @version      2021.4.25.2
+// @version      2021.9.25
 // @description  Collapses work attributes when there are too many. Workaround for MBS-11535/MBS-11537.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


ROpdebee:
@kellnerd You're more knowledgeable about these stanzas than I am, do you know of a better way to do this?

mb_collapse_work_attributes.user.js
```diff
@@ -9,8 +9,8 @@
 // @updateURL    https://raw.github.com/ROpdebee/mb-userscripts/main/mb_collapse_work_attributes.user.js
 // @match        *://*.musicbrainz.org/work/*
 // @match        *://musicbrainz.org/work/*
-// @match        *://*.musicbrainz.org/artist/*/works
-// @match        *://musicbrainz.org/artist/*/works
+// @match        *://*.musicbrainz.org/artist/*/works*
+// @match        *://musicbrainz.org/artist/*/works*
```
kellnerd, line 13:
```suggestion
// @match        *://*.musicbrainz.org/artist/*/works
// @match        *://*.musicbrainz.org/artist/*/works?*
```
This might be the safer way to do this with match patterns but I still have to test this with VM and TM.

ROpdebee: Works on my end, will update and force-push. Also going to simplify the other `@match` URLs to use *.mb instead of duplicating.

Btw, if you're looking for an artist to test on, Dolly Parton is one.

kellnerd: Yes, I also tested this with Dolly Parton right now ;)

kellnerd (approved):

PR #64, opened 2021-09-25
Support other languages on ISWCNet

### LINKED ISSUE #18: Support other languages on ISWCNet
Hi, I'm trying to use your script, (chrome latest version with Tampermonkey) and when I click "copy work code" on ISWC, it doesnt get any info at all and when I tried to paste with the button it always report "Log - No data found. Did you copy anything?" Is there something I don't do right, sorry I might not be an Issue, probably an error on my part.

Also, I've tried with Firefox (greasemonky) and Vivaldi (violentmonkey)... always the same issue

ROpdebee:
Add support for languages other than English on ISWCNet.

When solving the captcha, ISWCNet presents you with a language choice. The language code gets stored into session storage under the name `culture`. That item is used to request `https://iswcnet.cisac.org/configuration/GetLocalizedStrings` which returns the translations for the given "culture" (i.e., language). The results are JSON-serialized and stored into local storage under the key `strings`.

We're using that local storage item to translate the necessary text so that we can properly extract information on non-English ISWCNet pages. Since class names seem auto-generated and may be subject to change, matching DOM elements works through scanning for elements with a certain text, and therefore we need the correct translations. Moreover, the ID of some elements depends on a translated string as well (notably the IDs of table cells).

Fixes #18.

### DIFF
--- a/mb_bulk_copy_work_codes.user.js
+++ b/mb_bulk_copy_work_codes.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Bulk copy-paste work codes
-// @version      2021.5.27.2
+// @version      2021.9.25
 // @description  Copy work identifiers from various online repertoires and paste them into MB works with ease.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


mb_bulk_copy_work_codes.user.js
```diff
@@ -579,16 +579,43 @@ function storeData(source, iswcs, codes, title) {
...
+let translateStrings = (function() {
+    let strings;
+    const stringsDefaults = {
+        AGENCY_NAME_FIELD: 'Agency Name',
+        AGENCY_WORK_CODES: 'Agency Work Codes',
+        AGENCY_WORK_CODE_FIELD: 'Agency Work Code',
+        ARCHIVED_ISWCS: 'Archived ISWCs',
+        ORIGINAL_TITLE_FIELD: 'Original Title',
+        PREFERRED_ISWC_FIELD: 'Preferred ISWC',
+    };
```
ROpdebee, line 591: These are the strings which used to be hard-coded. We're using these as defaults in case `strings` isn't present in local storage for any reason whatsoever.

kellnerd (approved):
The copy button works for me on French ISWCNet page

PR #66, opened 2021-09-26
Qobuz improvements

### LINKED ISSUE #55: Qobuz: Goodies
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/26
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/28

### LINKED ISSUE #56: Qobuz: Support `open.qobuz.com` links
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/28

ROpdebee:
* Support `open.qobuz.com` links (fixes #56)
* Fetch goodies where available (fixes #55)

Goodies are fetched through the API with the App ID used on the `open.qobuz.com` webpages. I'm not sure whether this App ID changes often. For simplicity's sake, we're currently assuming that it doesn't, but if it does, it should be feasible to extract it automatically in the future.

This doesn't support back covers yet, as I haven't found any example of a release with a back cover. The field is always `null` in the API response.

Note also that if the API ever refuses to return a result, the whole thing aborts. We could fallback to the URL rewriting approach used previously to at least get the cover, but then we risk not noticing that goodies are missing. So instead we refuse to fill anything and hope that if anything goes wrong, someone will tell me. Ideally we'd have a better way to report errors so we could fall back on URL rewriting and persistently display the error on the page so someone can report.


ROpdebee:
Also updated the script locations for the new build process introduced in #45

src/mb_enhanced_cover_art_uploads/providers/qobuz.ts
```diff
@@ -1,29 +1,126 @@
...
+// lead to issues with URLs like this:
+// https://open.qobuz.com/album/1234567890/related
+// Not sure if such URLs would ever occur, but using a single regexp could
+// lead to `related` being matched as the ID and the actual ID as the title.
+const WWW_ID_MATCH_REGEX = /\/album\/[^/]+\/([A-Za-z0-9]+)(?:\/|$)/;
+const OPEN_ID_MATCH_REGEX = /\/album\/([A-Za-z0-9]+)(?:\/|$)/;
+// Assuming this doesn't change often. If it does, we might have to extract it
+// from the JS code loaded on open.qobuz.com, but for simplicity's sake, let's
+// just use a constant app ID first.
+const QOBUZ_APP_ID = 712109809;
```
kellnerd, line 19: I can confirm that this ID is not somehow user-dependent, I could extract the same number. Other than that there is not much to test for me, since I only know the one release with a booklet that you have probably already used for your test.

ROpdebee: Since the ID is embedded in the `main.js` file, I'm assuming it's neither user-dependent or time-dependent. But if they update their JS at some point, the App ID might change with it. Time will have to tell, which is why I'm forcing the provider to fail whenever the API doesn't return a valid response.

ROpdebee:
@kellnerd OK to merge?

kellnerd (approved):
Yes, code LGTM but I did not test a built version of the userscript.

ROpdebee:
I need to figure out how to get a built version of a PR provided automatically. Might be an option to have the GH Actions bot comment on PRs with a link to the built userscript?

Automated tests would also be a nicety to have, I'm slowly working towards that now.

src/mb_enhanced_cover_art_uploads/providers/qobuz.ts
ROpdebee: That ID has been present in their `main.js` since 2017, according to [WBM](https://web.archive.org/web/20171230102128/http://open.qobuz.com/resources/0.0.28/js/main.js). So I doubt it'll stop working any time soon :)

PR #68, opened 2021-09-26
Add tests

ROpdebee:
WIP, but sharing early, especially because this PR might grow to quite a large size.

~Run tests using `npx jest` (TODO: add a `package.json` script).~
Run tests using `npm run test` or `npx jest` or `npm t`.

The order of jobs in CI is currently a bit backwards (test -> build -> release makes little sense), but it works for now. Once tests that are ran in an actual browser are added, that order has to adapted.


ROpdebee:
Every time I add tests to a module, I find stuff to refactor, so I'm half-tempted to merge this already and then work on tests and refactoring simultaneously in smaller PRs.

For example, the `gmxhr` refactoring (#69) already conflicts with the tests added here...

kellnerd:
You don't have to create a separate branch for each change just for the purpose of easing reviews, do whatever feels most convenient for you. As long as there are separate commits for the different changes, I have no problem with a huge "refactoring & tests" PR. However, I think that I won't have time to review any PR in detail until the end of the week.

ROpdebee:
I decided against refactoring `qs`/`qsa`/`qsMaybe` because both `qs` and `qsa` do useful things (asserting existence and converting to a real array, respectively). `qsMaybe` does nothing but wrap `querySelector` though, so it's a bit useless, but IMO it makes sense to keep it for the sake of consistency.

ROpdebee:
I've done some experimentation, it appears that we'll run into quite a lot of problems if we try to test files containing JSX. We'd need to run them through babel first, but we can't do that with `nativejsx` currently. I don't feel like changing the way we handle JSX at the moment. I _could_ write a babel plugin, but that's of course not going to be straightforward and I doubt it's worth it. On top of all that, the SCSS imports will be quite the challenge to deal with too.

Instead, I'd propose to split off the UI elements into separate TSX files whose sole responsibility will be UI stuff. I should be able to mock those imports in the unit tests, and besides, it's good software design to separate UI from logic anyway :)

ROpdebee:
@kellnerd Marking this one as ready, as the `lib` and `build` stuff is fully covered with these tests, where possible*. I'm going to branch off to a separate PR to test the ECAU script source, since this one has already grown sufficiently.

Feel free to take a look through some of the changes on your own pace if you'd like, merging these changes isn't urgent. I'll likely go over the history once more myself soon-ish. Just let me know if you intend on reviewing or if I should just merge whenever I complete my double-checking pass. I'm fine with either. :)

*In terms of `build`, I'm only really testing the userscript metadata generator, not anything else. I'd rather not touch the build scripts too much, since I can't fully remember how they work 😅 Large swaths of its code are very difficult to properly test, and most (if not all) of that code is exercised by the CI build anyway. That of course doesn't verify that the scripts run properly when built, but I intend on adding real E2E browser tests at some point in the near future anyway, after unit testing the script code itself.

kellnerd:
I will have a closer look over the weekend and might leave you some comments 😄

tests/build/plugin-userscript.test.ts
```diff
@@ -0,0 +1,298 @@
+import fs from 'fs';
+
+// @ts-expect-error rewired
+import { __get__ } from '../../build/plugin-userscript';
+
+
+describe('git URLs', () => {
+    const GitURLs = __get__('GitURLs');
```
kellnerd, line 8: Is this a work-around to not have to export the class because the export would only be needed by tests and is not expected to be imported from anywhere else?

tests/build/plugin-userscript.test.ts
```diff
@@ -0,0 +1,298 @@
...
+        it('inserts license', async () => {
+            packageReaderMock.mockResolvedValue(JSON.stringify({ ...packageJsonWithAuthor, license: 'MIT' }));
+
+            await expect(metaGen.insertDefaultMetadata({}))
+                .resolves
+                .toMatchObject({ license: 'MIT' });
+        });
+
+        it('inserts namespace', async () => {
+            packageReaderMock.mockResolvedValue(JSON.stringify({ ...packageJsonWithAuthor }));
```
kellnerd, line 227:
```suggestion
            packageReaderMock.mockResolvedValue(JSON.stringify(packageJsonWithAuthor));
```

tests/build/plugin-userscript.test.ts
```diff
@@ -0,0 +1,298 @@
...
+
+            await expect(metaGen.insertDefaultMetadata({}))
+                .resolves
+                .toMatchObject({ namespace: 'https://github.com/ROpdebee/mb-userscripts' });
+        });
+
+        it('inserts homepage URL', async () => {
+            packageReaderMock.mockResolvedValue(JSON.stringify({ ...packageJsonWithAuthor }));
```
kellnerd, line 235:
```suggestion
            packageReaderMock.mockResolvedValue(JSON.stringify(packageJsonWithAuthor));
```

tests/build/plugin-userscript.test.ts
```diff
@@ -0,0 +1,298 @@
...
+
+            await expect(metaGen.insertDefaultMetadata({}))
+                .resolves
+                .toMatchObject({ homepageURL: 'https://github.com/ROpdebee/mb-userscripts' });
+        });
+
+        it('inserts issues URL from git repo', async () => {
+            packageReaderMock.mockResolvedValue(JSON.stringify({ ...packageJsonWithAuthor }));
```
kellnerd, line 243:
```suggestion
            packageReaderMock.mockResolvedValue(JSON.stringify(packageJsonWithAuthor));
```

tests/lib/MB/EditNote.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(textarea.value.trim()).toBe('info line');
+    });
+
+    it('retains pre-existing lines', () => {
+        textarea.value = 'preamble';
+        const editNote = new EditNote('footer');
+        editNote.addExtraInfo('info line');
+
+        expect(textarea.value.trim()).toBe('preamble\ninfo line');
+    });
```
kellnerd, line 57: This looks almost identical to 'retains previous content', the only differences seem to be the order of the first two statements and whether `trim()` is called or not (which should have no effect here).

kellnerd (approved):
Looks good to me, I have run the tests locally with `jest --verbose` and `jest --coverage`. Lots of tests which should cover all the code in the `src/lib` if I can trust jest. I definitely would have been too lazy to write most of them ("this method is unlikely to break, that one too..." 😅).

P.S. Just to make it clear, while I know the theory about unit testing (in JS and Python), I have never actually written tests using a framework since my projects are usually to small for me to bother with, or die before they reach a state which would require proper testing framework.

package.json
```diff
@@ -3,11 +3,12 @@
   "version": "1.0.0",
   "description": "Build package for MB userscripts",
   "scripts": {
-    "build": "NODE_ENV=production node --unhandled-rejections=strict --loader ts-node/esm rollup.ts",
-    "build-dev": "node --unhandled-rejections=strict --loader ts-node/esm rollup.ts",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "build": "NODE_ENV=production node --unhandled-rejections=strict --loader ts-node/esm build/rollup.ts",
+    "build-dev": "node --unhandled-rejections=strict --loader ts-node/esm build/rollup.ts",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx --max-warnings=0",
```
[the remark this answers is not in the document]
ROpdebee: In retrospect, it doesn't :) But AFAIK it does no harm setting them to `error` and reverting them back to `warning` would make the git history even more messed up that it already is with these parallel PRs :p

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 dist/
 .DS_Store
+coverage/
```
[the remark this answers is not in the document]
ROpdebee: Yeah I should add a package script for coverage, maybe integrate some external service to visualise coverage after CI too.

tests/lib/util/metadata.test.ts
```diff
@@ -0,0 +1,9 @@
+import { transformMBMatchURL } from '@lib/util/metadata';
+
+
+describe('transforming match URLs', () => {
+    it('should prefix MB domain', () => {
+        expect(transformMBMatchURL('release/add'))
+            .toBe('*://*.musicbrainz.org/release/add');
```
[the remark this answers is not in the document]
ROpdebee: That would definitely be a good idea. Since this is a bit tricky, I'll reference this comment in a new issue and put my thoughts there, so that it doesn't get lost in this PR.

tests/build/plugin-userscript.test.ts
ROpdebee: Exactly. Perhaps a bit overkill though.

tests/build/plugin-userscript.test.ts
ROpdebee: 🤦‍♂️

tests/lib/MB/EditNote.test.ts
ROpdebee: They test different things. One tests whether it retains content that was already in the edit note before the instance was created, the other tests whether it retains content added to the edit not after the instance was created. I have to admit that it took me a couple of minutes to figure out the difference too, so I reordered the cases to put these closer together and renamed them to make the difference more noticeable.

ROpdebee:
> I definitely would have been too lazy to write most of them ("this method is unlikely to break, that one too..." 😅).

I always tell students that 100% code coverage doesn't mean much more than e.g. 90% code coverage. Your tests can achieve 100% coverage, but if you've got code like
```
function abs(x) { return x; }
```
and you only test `abs(1)`, `abs(0)` and `abs(100)`, you'll get 100% code coverage but your code will still do wrong things on `abs(-1)`.

Nevertheless, 100% code coverage looks nice :) I use coverage mainly as a guidance to check what I might have forgotten to test, and if everything is green now in the coverage overview, and next week I make a change and all of a sudden it's not green anymore, I'll immediately know that certain parts need to be tested again. But I'm not going to go out of my way to cover some statement which is nearly impossible, I'll just add an `/* istanbul ignore next */` comment before them (istanbul being the coverage provider) if I'm reasonably sure that 1) that statement is correct or 2) there will be another way to fail the deployment if the statement isn't.

Thanks for the review!

tests/lib/util/metadata.test.ts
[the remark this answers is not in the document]
kellnerd: Thanks, I should have done that myself right away. I guess I lean towards leaving comments like this one buried somewhere in the code instead of creating a more traceable issue 🙄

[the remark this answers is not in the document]
ROpdebee: Don't beat yourself up over it :)

PR #69, opened 2021-09-26
Refactor `gmxhr`

ROpdebee:
Little bit of refactoring because I dislike the interface of `GM_xmlhttpRequest` and since we're using a wrapper, we can do better.

* Make the `method` parameter optional, default to `GET` since that'll be used most often.
* Move the `url` to a separate parameter, since it's always required anyway, whereas other options might not be. Also accept `URL` instances.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/index.tsx
+++ b/src/mb_enhanced_cover_art_uploads/index.tsx
@@ -277,8 +277,7 @@ class ImageImporter {
     }
 
     async #fetchImage(url: URL, fileName: string, headers: Record<string, unknown> = {}): Promise<FetchResult> {
-        const resp = await gmxhr({
-            url: url.href,
+        const resp = await gmxhr(url, {
             responseType: 'blob',
             headers: headers,
         });
--- a/src/mb_enhanced_cover_art_uploads/index.tsx
+++ b/src/mb_enhanced_cover_art_uploads/index.tsx
@@ -279,7 +279,6 @@ class ImageImporter {
     async #fetchImage(url: URL, fileName: string, headers: Record<string, unknown> = {}): Promise<FetchResult> {
         const resp = await gmxhr({
             url: url.href,
-            method: 'GET',
             responseType: 'blob',
             headers: headers,
         });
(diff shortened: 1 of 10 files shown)


kellnerd:
> I dislike the interface of GM_xmlhttpRequest

Me too, I have simply used [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) for my userscripts so far. It works flawlessly in a current browser with ViolentMonkey (and probably also in TamperMonkey, not sure whether I had already tested that).

ROpdebee:
`fetch` is great but I don't think it'd work properly for most of these requests because of CORS :( I'm still using `fetch` for requests to `/ws/` and other MB pages though.

src/mb_enhanced_cover_art_uploads/index.tsx
```diff
@@ -11,6 +13,17 @@ import { ArtworkTypeIDs } from './providers/base';
...
+    url: string | URL
+    fileName: string
+
+    constructor(url: string | URL, fileName: string) {
+        super(`${fileName} has an unsupported file type`);
+        this.url = url;
+        this.fileName = fileName;
+    }
+}
+
```
kellnerd, line 26: I don't think this class should be in the main module of the userscript, a separate module for custom errors might be a cleaner solution. Not sure if the `ResponseError` classes in the XHR module should be moved as well, they rather suit their module compard to the `BadFileTypeError`.

kellnerd (commented):
Thanks, the new interface is much better, I only have one minor comment.

Regarding the potential CORS issues with `fetch`, I guess I just had luck so far that I was only fetching MB resources and calling the Discogs API (which has the `Access-Control-Allow-Origin: *` header) from userscripts which are running on MB.

src/mb_enhanced_cover_art_uploads/index.tsx
ROpdebee: I agree, but I'll be doing much more refactoring in #72, so I say we address that there :)

kellnerd (approved):

src/mb_enhanced_cover_art_uploads/index.tsx
ROpdebee: Actually, we could just remove this custom error subclass altogether since we're never actually using it specifically, and just throw a normal `Error`. We're only using the `message` field anyway, and I doubt we'll ever need to do an `instanceof` check on these.

The `ResponseError` classes in XHR have more purpose. Even though we're not currently using them in ECAU, I feel like they could still be useful in other scripts.

PR #70, opened 2021-09-27
Fix duplicate image detection on error

ROpdebee:
Noticed this while testing for #69. We were adding the URL to the done set before it was actually done, and this prevents retrying the same URL if it failed previously. Fixing by adding the URLs to the done set after the image was queued.


kellnerd (approved):
And I have not even noticed this logical error while I was fixing the duplicate image check for the first time in #39 😅

Issue #81, opened 2021-10-16
Improve status banner GUI


ROpdebee:
The current status banner of ECAU only shows the most recent message. It should be replaced by a nicer GUI which retains the previous messages as well, similar to the ones in the work code paster.

PR #82, opened 2021-10-16
Improve safety when URLs redirect

### LINKED ISSUE #79: URL redirect safety
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/37

> Warn about redirects, maybe indicate this in the note too
> Especially [Bandcamp is known to be a fustercluck](https://community.metabrainz.org/t/bandcamp-release-date-discrepancies/401016/18) and one can never be careful enough:
>     https://musicbrainz.org/edit/83531854
>     [Edit #83531532 - MusicBrainz](https://beta.musicbrainz.org/edit/83531532)

Marking as high priority as it has led to wrong edits before.

ROpdebee:
Highlights:
* Each provider now checks whether any redirects occurred and if so, whether the redirect is "safe" (i.e., redirects to the same release). If the redirect is unsafe, the extraction process is aborted.
   * For providers which load the HTML of the provider page (e.g., Bandcamp, Apple Music, Amazon), we check whether the page is redirected and whether the unique ID of the page URL is the same before and after the redirect.
   * For providers which use an API (Discogs, Qobuz, etc), we check whether the ID returned in the API response is the same as the one we requested, just to be sure.
* When a URL pointing directly to an image is redirecting, a warning is emitted and that redirection is indicated in the edit note.

Known issues:
* When provider extraction is aborted, the script logs an error asking the user to double-check the redirection and use the redirected URL itself if it's still correct. However, that message is too long, leading to bad positioning of the DOM elements. Could be easily fixed with some CSS, but since we'll add an improved status banner GUI at some point in the future (#81), I decided against torturing myself with CSS today. Those redirections should be rare enough anyway, so it won't lead to issues too often.
* The warning emitted when a direct image link is redirecting isn't going to be visible to the user for very long, since it'll be overwritten by a success message quite quickly. Again, this should get fixed when #81 is implemented.

Closes #79.
Depends on #72.
~Bumping version for release.~

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
+++ b/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
@@ -3,7 +3,6 @@ import { qs, qsa, qsMaybe } from '@lib/util/dom';
 import type { CoverArt } from './base';
 import { ArtworkTypeIDs, CoverArtProvider } from './base';
 
-const ID_REGEX = /\/(?:gp\/product|dp)\/([A-Za-z0-9]{10})(?:\/|$)/;
 const PLACEHOLDER_IMG_REGEX = /01RmK(?:\+|%2B)J4pJL/;
 
 export class AmazonProvider extends CoverArtProvider {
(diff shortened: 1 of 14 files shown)


ROpdebee:
Dropping the version bump again, we'll merge this after #84 so that we can test whether CD deploys correctly 😊

PR #84, opened 2021-10-17
Continuous deployment

ROpdebee:
I originally intended to only enable CD after E2E tests were added, but since there's now 2 PRs open which could both be released as a new version and  I didn't want to run into conflicts by bumping the version in both PRs separately, I'm adding it now.

Quick summary of the changes:
- Workflows on PRs will remain the same.
- Workflows on pushes to main (i.e. after PR merges) are removed and replaced by workflows that run after a PR is merged. The reason is that this allows us to more easily figure out the PR associated with the change. We'll need this to report deployment status on the PR, and also to report the PR in the deployment commit message.
- Versions are now automatically generated by bumping from the version extracted from the previous deployed version. So `meta.ts` files don't contain versions anymore.
- Deployments are skipped if 1) building or testing fails; 2) there is no change between the old deployed version and the built version after the PR is merged (differences in the versions are taken into account here); or 3) a "skip cd" label is set on the PR.
- There's now a scheduled nightly workflow which builds and tests userscripts, to replace the workflow that usually ran on pushes to `main`. That scheduled workflow will also create an issue if it fails (at most one issue will be open at any given time, if a run fails and an issue already exists and is open, the existing issue will be edited).
- `npm run build` and `npm run build-dev` now generate a version themselves, but this should only really be used for local development usage or in CI to check that the scripts can be built correctly.

### DIFF
--- a/build/deploy.ts
+++ b/build/deploy.ts
@@ -1,5 +1,6 @@
 import simpleGit from 'simple-git';
 import fs from 'fs/promises';
+import path from 'path';
 import { getPreviousReleaseVersion, incrementVersion } from './versions.js';
 import { buildUserscript } from './rollup.js';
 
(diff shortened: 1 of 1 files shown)


ROpdebee:
Merging this without review because it's blocking some other PRs and the easiest way to make sure it's working properly is to actually push it :) Apologies! Feel free to leave any remarks and I'll do the fixes afterwards.

PR #86, opened 2021-10-17
Extract Amazon image URLs from the page's embedded JavaScript

### LINKED ISSUE #85: Amazon provider does not return the highest resolution
https://musicbrainz.org/release/9f1548cd-5bde-4fb6-9e11-921e9775a38e/add-cover-art returns only 500px images from Amazon although there are larger ones:

> https://www.amazon.com/gp/product/B07C5DXCQ4
>  * https://m.media-amazon.com/images/I/415ge6BXIIL.jpg
>    Maximised from https://m.media-amazon.com/images/I/415ge6BXIIL._SS40_.jpg
>  * https://m.media-amazon.com/images/I/41LvFX+GzpL.jpg
>    Maximised from https://m.media-amazon.com/images/I/41LvFX+GzpL._SS40_.jpg
> –
> MB: Enhanced Cover Art Uploads 2021.9.26
> https://github.com/ROpdebee/mb-userscripts

It should maximise the image itself (which is already 1500px or 1200px) and not the thumbnails, in this case that approach would return 2400px images:

> https://www.amazon.com/gp/product/B07C5DXCQ4
>  * https://m.media-amazon.com/images/I/81CaO+EAEKL.jpg
>    Maximised from https://m.media-amazon.com/images/I/81CaO+EAEKL._SL1500_.jpg
>  * https://m.media-amazon.com/images/I/81yqjMnCpmL.jpg
>    Maximised from https://m.media-amazon.com/images/I/81yqjMnCpmL._SL1500_.jpg

The problem is that these have a different identifier than the thumbnails and only one image is shown at a time. `const imgs = qsa<HTMLImageElement>('li.image img', pageDom);` instead of `const imgs = qsa<HTMLImageElement>('#altImages img', pageDom);` would give us all the images with the correct identifier and a `._SY450_` suffix, but only after the other ones have been opened/viewed by the user 😞

We probably have to extract the image URLs from the page's embedded JavaScript.

kellnerd:
Using the thumbnails which are shown on the page has two disadvantages:
- Pages show only a limited number of thumbnails, not all of them.
- Maximised versions of these might still not be the largest available.

Unfortunately the most reliable way (so far) to get all images in their highest resolution, is to extract an object from the page's embedded JavaScript. The advantage of this approach is that we can also extract artwork types from there now (at least 'Front' and 'Back', maybe others).

Closes #85.
Depends on #82 (~and needs to be rebased~).

P.S. This is a non-issue for digital media releases, so I have not touched that code. And I kept the old thumbnail extraction code as a fallback (in case the regex does not find a match in the page's source).


ROpdebee:
Take your time, I'm in favour of postponing the merge until after #115 to check whether the new workflows work properly 😄

PR #87, opened 2021-10-17
docs: prettify README

ROpdebee:
* Add various status badges. Some of them might not work yet as
  they depend on changes that are added in open PRs.
* Change source and install buttons for userscripts. Buttons now
  also include the current version of the script. For old scripts,
  that version is static, but for new scripts that use the new
  CD pipeline, the version will by dynamically retrieved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,22 +13,22 @@ Collection of userscripts for MusicBrainz.
 
 Blinds editor and voter details before your votes are cast.
 
-[![Source](https://img.shields.io/badge/source-main-lightgrey?style=for-the-badge&logo=github)](mb_blind_votes.user.js)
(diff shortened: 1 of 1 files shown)


ROpdebee:
I'm not sure. IMO it'd make sense to remove those Source buttons altogether, I don't really see the point in them.

kellnerd:
I guess they might be a relict from the past when userscipt managers did not show a preview anyway?
One case where they might be useful is for new users which have not already installed a userscript manager, because through this button they could still view the code with syntax highlighting (without having to find the corresponding file) and not just the raw plain text rendering behind the install button, but that is a relatively minor target audience.

Personally I use them relatively often if I want to see the code and browse the git history. But for automatically built userscripts this is almost pointless. It might make more sense to link the source directory? The label could be the branch name instead of the version in that case. And it would make sense for the install button to be the first one since it is more commonly used.

PR #92, opened 2021-10-17
fix(ci): remove bad check in deployment workflow

ROpdebee:
We previously checked whether `github.ref` pointed to the main branch,
but this is always false in workflows triggered by PRs (unless the source
branch is the main branch), because `github.ref` points to the ref
of the source branch. Since this workflow is explicitly only enabled
on PRs that have `main` as the target, the check is redundant.


ROpdebee:
Not skipping CI on this one, because it should've deployed on #82, but that failed due to this bug.

PR #141, opened 2021-10-20
test(caa upload): make sure Tidal provider does not extract video

ROpdebee:
I bumped into a Tidal release where the front cover is actually a video. Since video covers aren't supported on CAA yet (see [CAA-132](https://tickets.metabrainz.org/browse/CAA-132)), we should make sure that we grab the static image rather than the cover.

It turns out that this already works properly in the code, but I'm adding a test case to make this explicit and prevent regressions in the future.


ROpdebee:
/deploy-preview

For future reference: It seems like the URL to the video cover is present in the JSON response too, in case we want to add support for video covers in the future.

PR #149, opened 2021-10-21
fix(caa edits): prevent DOM-based code injection

ROpdebee:
The way we were splitting the types TD into multiple spans would
make it possible to inject arbitrary HTML. I can't immediately
see how this were exploitable in practice besides someone being
able to maliciously update MB's translations. As good practice,
and for safety, let's just fix it anyway. This shouldn't be
exploitable anymore since `textContent` doesn't parse HTML tags.

See https://github.com/ROpdebee/mb-userscripts/security/code-scanning/8


ROpdebee:
I'm quite surprised to see that it's been half a year since Supercharged CAA Edits was updated...

PR #151, opened 2021-10-21
fix(caa upload): match underscores in a-tasket's self_id query param

kellnerd:
The seeder previously returned incomplete cached URLs for a-tasket pages, i.e. it only returned only `cached=<barcode>-d` instead of `cached=<barcode>-d_<deezerId>-s_<spotifyId-i_<appleId>`.

A cleaner solution would be the usage of `URLSearchParams` instead of regular expressions. Is/Was there any reason no to use them?

*Edit*: In any case, that's it from me for today, feel free to release this as a hotfix (we can do the cleaner solution later) or push directly to this branch if you want.


ROpdebee:
If you're feeling daring, those `URLSearchParams` could probably also heavily simplify the decoding in `seeding/providers.ts`. Perhaps even the encoding (if it does that too). 😄

PR #195, opened 2021-10-29
fix(caa upload): Remove duplicate and placeholder images

### LINKED ISSUE #172: Filter out placeholder images
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/41

Related to #171.

### LINKED ISSUE #174: Bandcamp: Duplicate track images
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/41

We need to do extra deduplicate based on image data. Doesn't look like they return ETag or Digest headers either.

ROpdebee:
* Remove duplicate Bandcamp track images
  * We first use the thumbnails to remove duplicates, as is done for Soundcloud
  * Added a postprocessing hook to the providers that allows a provider to change the images to be queued after they are fetched.
  * Using the contents of the full-sized fetched image, we do another round of deduplication, since for Bandcamp, it seems that thumbnails can still differ even though the original is the same, see #174.
  * Some refactoring to move track image merging to a base class.
  * It won't be 100% reliable, but I don't think we'll be able to do much better. I've checked the releases chaban posted in the forums and all of them are behaving as they should. There are a couple of them where it looks like it grabbed the same image twice, but on closer inspection you'll notice that they're slightly different (diff brightness in one case, some sort of 1px white border on another, etc.)
  * Closes #174.

* Since there's now the postprocessing hook, I decided to also address the placeholder images for Deezer and 7digital in this one:
  * Deezer image content signatures are compared to that of the placeholder cover, since I'm not sure whether it'll always be the same URL.
  * We check 7digital fetched images' final URL (after redirection) because 404 images are redirected to the placeholder, which always seems to have the same URL.
  * Closes #172.

I'm not entirely satisfied with this. Style and architectural suggestions most welcome.


src/mb_enhanced_cover_art_uploads/providers/base.ts
```diff
@@ -194,3 +180,129 @@ export abstract class HeadMetaPropertyProvider extends CoverArtProvider {
...
+                    const digest = await this.#urlToDigest(coverUrl);
+                    // This will overwrite any previous entry if the digest is the same.
+                    // However, that's not a problem, since we're intentionally deduping
+                    // images with the same payload anyway. It doesn't matter which digest
+                    // we use in the end, all of those digests return the same data.
+                    dataToOriginal.set(digest, coverUrl);
+                    return trackCovers.map((cover) => {
+                        return {
+                            ...cover,
+                            url: digest,
```
kellnerd, line 273: This assignment is a bit confusing, a comment to indicate that this happens purely to satisfy the requirements of `#groupIdenticalImages()` would clarify this.

kellnerd (commented):
I have to admit that I still haven't tested these changes, so I will at least submit my comments which have been pending for two days now, I even deleted some of them which were no longer valid (or did no longer make sense to me 😅).

Personally I think this is probably too much complexity for rather rare use cases. Most releases won't have duplicates or not even multiple images (the common case), so calculating the hash of all fetched images is usually a wasted effort. In my opinion it's still the reponsibility of the user to check whether there are duplicate, inferior or otherwise redundant images before submitting. But I won't complain (...well, more than I already have) until I have tested the performance impact for a large release with lots of unique images (e.g. Discogs).

tests/mb_enhanced_cover_art_uploads/providers/7digital.test.ts
```diff
@@ -48,4 +48,24 @@ describe('7digital provider', () => {
         await expect(provider.findImages(new URL('https://de.7digital.com/artist/mnek/release/language-elicit-8354116')))
             .rejects.toThrowWithMessage(Error, 'HTTP error 404: Not Found');
     });
+
+    it('filters out placeholder images', async () => {
```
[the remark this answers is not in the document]
ROpdebee: I noticed this in #209 too, and I forgot that these tests were introduced here.

ROpdebee:
> Personally I think this is probably too much complexity for rather rare use cases.

I kind of have to agree, which is unfortunate since it's already implemented 😂 I sort of hate the implementation though, it's a massive hack (especially the track comment merging…). I'm sort of in favour of dropping the whole "deduplicate by full image digest" part, since IIRC it was just one of all of those Bandcamp releases chaban linked that this was fixing.

I think deduplicating by URL and thumbnail content is still fine (the latter is necessary for Soundcloud anyway, might as well reuse it in Bandcamp), using the digests for the thumbnails seems to be okay too since it won't be calculated _that_ often. The post-processing step is fine too to remove the placeholder images and perhaps do some other fixups in the future. But I despise the implementation of the full-size deduplication.

tests/mb_enhanced_cover_art_uploads/providers/deezer.test.ts
```diff
@@ -51,4 +51,26 @@ describe('deezer provider', () => {
...
+            digest: '2a16c47b2769e6f8414c3f8e39333b46f9b61a766e1dfccc2b814767d3b662cb',
+        }]];
+        // @ts-expect-error: Lazy
+        const afterFetch = await provider.postprocessImages(fetchResults);
+
+        expect(afterFetch).toBeEmpty();
+    });
+
+    it('does not filter out legit images', async () => {
+        // For https://www.deezer.com/de/album/673547
```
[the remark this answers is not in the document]
ROpdebee: Marking as resolved since these tests were removed.

ROpdebee:
I've removed the digest-based post-processing deduplication, I'll double check how this affects the Bandcamp releases posted by chaban. Also removed the digest calculation entirely (except for thumbnail-based deduplication) and switched Deezer to filter placeholders by URL, since the URL already contains an MD5 sum and it's the same URL for all releases with a placeholder, apparently.

src/mb_enhanced_cover_art_uploads/providers/base.ts
ROpdebee: Refactored the implementation and sprinkled in some more comments. It should be more elegant and clear now.

ROpdebee:
/deploy-preview to test Bandcamp please

I feel like the current implementation is a lot better than the original one. There's one branch which is not covered by the tests, but it is covered in the tests introduced in #209. I wanted to move those tests here, but it depends on a lot of the abstractions introduced there, so for simplicity's sake, let's leave the branch uncovered for now and fix that in the test refactoring.

ROpdebee:
OH. MY. F***ING. GOD.

68a616b962dd505aec93e359074b2aca4b0acf5d

"We need to deduplicate by full-res content since sometimes the thumbnails may be different even though the originals are identical."
It turns out that wasn't true at all. I swear I actually calculated MD5 sums of these supposedly different thumbnails, but I guess I somehow still compared the wrong things? I'm not sure. Regardless, it turns out that the thumbnails of the release I complained of in #174 actually _are_ the same. The problem was that we were never deduplicating a track image with the main image if there was only one track image after URL deduplication.



I've tested chaban's releases with the version of the most recent commit and there are 0 duplicate images (except for the expected ones whose originals _actually_ differ).

src/mb_enhanced_cover_art_uploads/providers/base.ts
```diff
@@ -43,6 +46,15 @@ export abstract class CoverArtProvider {
      */
     abstract findImages(url: URL, onlyFront: boolean): Promise<CoverArt[]>
 
+    /**
+     * Postprocess the fetched images. By default, does nothing, however,
+     * subclasses can override this to e.g. filter out or merge images after
+     * they've been fetched.
+     */
+    postprocessImages(images: Array<[CoverArt, FetchedImage]>): Promise<FetchedImage[]> {
+        return Promise.resolve(images.map((res) => res[1]));
```
kellnerd, line 55:
```suggestion
    postprocessImages(images: FetchedImage[]): Promise<FetchedImage[]> {
        return Promise.resolve(images);
```
No longer in use since https://github.com/ROpdebee/mb-userscripts/pull/195/commits/6495e35fd9fd2cfcdcbfe2b96c9072703b24268a, I'm not even sure if it has ever been necessary to have access to the `CoverArt` object for that code to work (since `FetchedImage` also has the `comment` property which was used in the now removed code).
Simplifying this paramter also requires changes in all modules where this method is called or overwritten.

kellnerd (commented):
> https://github.com/ROpdebee/mb-userscripts/commit/68a616b962dd505aec93e359074b2aca4b0acf5d

Wow, it's kind of fascinating how such a simple mistake as `>` instead of `>=` caused all this hassle. When I reviewed that condition, it looked logical to skip deduplication if there is only a single image, I also didn't notice that it has been the count of track images (and not of all images).

The code is much easier to read now that most of the complicated code is gone (without any drawbacks as it seems). Now that the hashes are only calculated for track image deduplication, I can live with the performance impact, having to wait until all the track pages of a Bandcamp release have been fetched strikes harder.

src/mb_enhanced_cover_art_uploads/providers/base.ts
ROpdebee: Good catch!

kellnerd:
I just noticed that the URL input does not get cleared for me (Firefox 94.0, Violentmonkey) after the URL has been processed, I get a type error instead:

> Uncaught (in promise) TypeError: evt.currentTarget is null

It happens only for `case 29: if (evt.currentTarget.value === oldValue) {` while `case 2: oldValue = evt.currentTarget.value;` seems too work!?
I have included `log.debug(evt)` statements in both places and it is indeed logged as `currentTarget: null` in both places, but logging `oldValue` for case 2 works nevertheless...
Don't know if this is caused by the huge generated JS wrapper, the original code in `main.tsx` looks sensible to me.

Replacing `currentTarget` with `target` fixes the issue for me, but TypeScript is unhappy with that as `target` is not guaranteed to have a `value` property.

ROpdebee:
Ahh. Yes. That was accidentally introduced in another PR, but I'll fix it here (an E2E test might've been able to catch it earlier, maybe I should get started on those :innocent:).

See the note at [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget). It doesn't have anything to do with the async transpilation, I used the original code (minus the type annotations) for that callback and it still leads to the same error. The problem is that it's asynchronous code, so by the time that `case 29` thing is executed, the synchronous part is already done and the `currentTarget` has been reset. It's pretty much this:

```javascript
function handler(evt) {
  const oldValue = evt.currentTarget.value;
  setTimeout(() => {
    /* evt.currentTarget is now null */
    if (evt.currentTarget.value === oldValue) { /* ... */ }
  }, 100);
}

const evt = { currentTarget = 123 };
handler(evt);
evt.currentTarget = null;
```
The culprit isn't really the asynchronous code though. The culprit is mutable data. Oh, the joys of mutability... And perhaps sloppy API design. Sloppy API design, in JavaScript? I'd never expect that! :joy: (/s)

As for why TS complains about `target`, it makes sense: `target` is the element on which the event occurred, which could be a descendant of arbitary type (in this case it'll always be the `input` element though). `currentTarget` is the element on which the handler is installed (always the `input` element), but changes for the event as it bubbles up the DOM. That also sort of explains why it's `null` after the event fully bubbled, but `target` remains to be the element on which the event was triggered, which doesn't change.

kellnerd (approved):
Shame on me, I have only read the `Event` page on MDN and did not notice there was a separate page for `currentTarget` 😊

ROpdebee:
Thanks for the extra set of eyes on this, @kellnerd! Quality of this PR has improved massively IMO.

PR #196, opened 2021-10-30
fix(caa upload): prevent processing URL if it's already in progress

### LINKED ISSUE #193: ECAU cover art gets added multiple times
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/43?u=ropdebee

I'm going to guess it's because we only add the processed URLs after they've been enqueued, so if it somehow happens that the button is clicked a second time before the first click is fully processed, the images may get added twice. I need to experiment a bit more to figure this out though.

ROpdebee:
May close #193 

In older versions, spam-clicking the provider button will lead to the same URL being processed multiple times, and possibly enqueued multiple times too, depending on the interleaving of certain async operations. To prevent this, we're now keeping a set of URLs which are currently in progress, and skip all processing of a URL entirely if it's already being processed.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/index.ts
+++ b/src/mb_enhanced_cover_art_uploads/index.ts
@@ -24,10 +24,12 @@ class App {
     #note: EditNote;
     #fetcher: ImageFetcher;
     #ui: InputForm;
+    #urlsInProgress: Set<string>;
 
     constructor() {
         this.#note = EditNote.withFooterFromGMInfo();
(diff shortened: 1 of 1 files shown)


kellnerd (approved):
Helps against double clicking an import button (tested with the Discogs provider).
Previously this queued the images (or most of them) twice and failed to generate an edit note body (only the footer was there). With the changes it only queues the images once and successfully generates an edit note body (as it does for a single click).

PR #197, opened 2021-10-30
feat(caa upload): add VGMdb seeder

### LINKED ISSUE #62: VGMdb: Not all images are fetched
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/31
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/17

I was under the impression that vgmdb.info provided all the covers, that's evidently not true. There doesn't seem to be a way for us to grab the missing images without somehow logging in to an account, and adding account support it something I'd like to avoid altogether.

It appears the most we can do here is warn if images are missing, but for that, we need to know how many images a given album has in total. That's available through [this table](https://vgmdb.net/db/covers.php), but extracting that information would be inefficient since we need to traverse the table page-by-page. Binary search would work, but still requires a fair few requests in the worst case.

At the very least, the provider support table should be updated to make note of this.

ROpdebee:
Closes #62.

We can't extract all images through the VGMdb provider, so we'll add a VGMdb seeder on logged-in pages so that images that require logging in can be extracted from the page itself, and added through seeding parameters.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10028,7 +10028,7 @@
     },
     "node_modules/nativejsx": {
       "version": "4.3.0",
-      "resolved": "git+ssh://git@github.com/ROpdebee/nativejsx.git#6bd78f05bf858978ad8217372a37594fb2a53e86",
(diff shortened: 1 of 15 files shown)


kellnerd (commented):
Looking good, but I could not test it since I do not have an VGMdb account.

PR #198, opened 2021-10-30
feat(caa upload): add MusicBrainz provider

### LINKED ISSUE #30: New provider: MusicBrainz
I think there are use cases where it makes sense to copy over images from one release to another.

ROpdebee:
This one might come in handy to copy images from one release to
another. Uses the `index.json` file stored in the IA to extract
all information. Won't show up in the provider buttons though.

Closes #30.


src/mb_enhanced_cover_art_uploads/providers/musicbrainz.ts
```diff
@@ -0,0 +1,48 @@
...
+        image: string;
+    }>;
+}
+
+export class MusicBrainzProvider extends CoverArtProvider {
+    supportedDomains = ['*.musicbrainz.org'];
+    favicon = 'https://musicbrainz.org/static/images/favicons/favicon-32x32.png';
+    override allowButtons = false;
+    name = 'MusicBrainz';
+    urlRegex = /release\/([a-z0-9-]+)/;
```
kellnerd, line 21: Good that the regex is not too strict, as you usually want to copy images when you are visiting `/cover-art` of the source release and will likely copy that URL 👍

kellnerd (approved):
LGTM, I have tested it on beta.mb with a `/cover-art` URL and it worked as expected. One observation which I made is that CAA URLs will always redirect which leads to longer edit notes, but that is probably the desired behaviour nevertheless.

Fun fact: In case you accidentally paste the userscript itself into its own URL input, it succesfully extracts the images of two URL examples from the code comments 😂

ROpdebee:
> One observation which I made is that CAA URLs will always redirect which leads to longer edit notes, but that is probably the desired behaviour nevertheless.

Yeah it's a bit noisy, but I decided to keep it that way since A) the MB provider probably won't be used that much anyway and B) IA regularly moves stuff between different servers so the URL it redirects to now, might change at a later date.

PR #199, opened 2021-10-31
fix(caa upload): extract track covers for releases without main cover

ROpdebee:
The changes from #152 prevent extracting any covers if the main release does not have a cover. However, there could still be separate track covers which we can extract.

Other smaller fixes:
* Don't log an error if a track is known to have no covers.
* Don't log messages about track cover extraction if there is no tracklist (e.g. for `/track/` URLs).


kellnerd (approved):
🚢

PR #200, opened 2021-10-31
feat(caa upload): allow fetching only the front cover

### LINKED ISSUE #37: Add option to only fetch the first image
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/17

> An option to abort and/or add only front/first cover. Especially Discogs can have quite a lot of images.
To save time and accommodate for people who don’t have as much bandwidth (admittedly for Discogs the latter is not much of an issue right now :frowning:)

Adding the option to only fetch the first image (presumably the front cover) would be a good compromise, although aborting in the middle might lead to partial image sets.

ROpdebee:
Closes #37.

This might be useful to some people, especially for providers with track covers, so let's add the option.

The UI can be improved further, but I hate CSS and it looks passable, so 🤷‍♂️


ROpdebee:
/deploy-preview since resolving merge conflicts may have screwed stuff up

src/mb_enhanced_cover_art_uploads/fetch.ts
```diff
@@ -140,14 +143,31 @@ export class ImageFetcher {
...
-        this.#doneImages.add(url.href);
+        if (!hasMoreImages && !onlyFront) {
+            // Don't mark the whole provider URL as done if we haven't grabbed
+            // all images.
+            this.#doneImages.add(url.href);
+        }
+
+        if (hasMoreImages) {
+            LOGGER.warn(`Not all images were fetched: ${images.length - finalImages.length} covers were skipped.`);
+        }
```
kellnerd, line 154:
```suggestion
        if (!hasMoreImages) {
            // Don't mark the whole provider URL as done if we haven't grabbed
            // all images.
            this.#doneImages.add(url.href);
        } else {
            LOGGER.warn(`Not all images were fetched: ${images.length - finalImages.length} covers were skipped.`);
        }
```
If we added only the fornt and there are noe more images we can still mark the whole provider URL as done.

kellnerd (approved):
Works as expected, the UI is good enough (better than too many import buttons which are floating into the next row 😉).

src/mb_enhanced_cover_art_uploads/fetch.ts
ROpdebee: Seems fair. I originally split them up because Bandcamp/Soundcloud won't return track images if we're grabbing only the front cover, but realistically, if someone's using this option, they probably won't go back to grab the track images anyway.

ROpdebee:
> better than too many import buttons which are floating into the next row 😉

Do you happen to have a release for testing? Or better yet, do you want to take a stab at fixing it? I'm bad at CSS 😅

kellnerd:
> > better than too many import buttons which are floating into the next row 😉
>
> Do you happen to have a release for testing?

Oh, I probably haven't described this accurately. The import buttons or the buttons and the URL input are always floating into the next row for me (and it does not really matter how many import buttons the cover art page has), because I usually use two browser windows side by side when I am editing on MB.

image

The display looks a bit ugly then because MBS uses an empty label to align the button, but I'm not sure if there is a clean fix which the userscript could do, this should rather be changed on MBS side to use flex-boxes or something similar.

ROpdebee:
Ah, I see. I've also noticed that when there's a lot of buttons, the UI doesn't look that great either. Flexbox would probably offer a solution, and I'm not opposed to the userscript patching the MB HTML to make it look better (after all, it's not a problem if the userscript isn't running, so it should also be the script's responsibility of fixing it). However, layout and styling really isn't my strong suit, so I'd just get frustrated trying to fix it. PRs welcome 😄

Also, MB dark mode?!

kellnerd:
If I did not have so many (coding-related) things with higher priority on my to-do-list, I would probably try to fix this myself, but since it's just a minor UI annoyance, I can life with it for a while 😇

> Also, MB dark mode?!

Oh, it's a userstyle which I created a few months ago and have been using ever since. First it took me a while to get used to the dark theme, but by now it's the default light theme which looks strange to me 😂
It still has a few rough edges and should be refactored a bit (e.g. by using CSS variables for identical colours), but I have now finally read the Stylus docs and [published it](https://github.com/kellnerd/userstyles#musicbrainz-dark).

PR #202, opened 2021-10-31
style: disallow use of confusing window globals

ROpdebee:
This will hopefully prevent issues like the one from #201


kellnerd (approved):
Wow, that list contains a lot of globals I was not aware of 😨

ROpdebee:
> Wow, that list contains a lot of globals I was not aware of 😨

Yet it still feels like there's some that are missing...

PR #215, opened 2021-11-03
fix(caa upload): polyfill `array.at`

### LINKED ISSUE #214: [cover art uploader] Uncaught (in promise) TypeError: e.split(...).at is not a function
I am on https://musicbrainz.org/release/6f2681d8-3a9e-3b45-b957-badc73649f96 trying to upload newly linked Discogs front cover:

> Uncaught (in promise) TypeError: e.split(...).at is not a function
> add-cover-art line 1 > scriptElement:104572:29695

Same error if I copy/paste image URL.
Same in beta MBS.

I am on Linux Debian 10, Firefox 78.15.0esr (64 bits).

ROpdebee:
I though MB polyfilled this one for us, but apparently it doesn't.

I'm not sure how I feel about including a whole bunch of polyfill code just for that single method (it's not actually polyfilling anything but `array.at`, but just happens to include a bunch of other stuff for core-js internals I guess?). At the same time, I want the source code to be modern and not have to be limited to old ES versions.

For this method though, I'm not sure whether it's worth the bloat... But on the other hand, the script also runs on pages other than MB, where polyfills might not be available.

Possibly closes #214

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/index.ts
+++ b/src/mb_enhanced_cover_art_uploads/index.ts
@@ -8,6 +8,8 @@ import USERSCRIPT_NAME from 'consts:userscript-name';
 import DEBUG_MODE from 'consts:debug-mode';
 import { seederFactory } from './seeding';
 
+import 'core-js/features/array/at';
+
 import { App } from './App';
 
 LOGGER.configure({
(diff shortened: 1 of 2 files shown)


ROpdebee:
I'm wondering whether it would make sense to separate all the third-party stuff into a vendor file and `@require` it instead of embedding it into the script itself...

kellnerd (approved):
I think a separate vendor file would make sense once the other userscripts have been refactored and include the same third-party stuff.

PR #217, opened 2021-11-04
Handle images without file extension

### LINKED ISSUE #216: Handle Spotify images without file extension
It seems like Spotify no longer uses filename extensions for their cover art. The a-tisket seeder displays TLD and path as extension for Spotify images now (e.g. [CO/IMAGE/AB67616D0000B273B8FFC252C4575933EE3BAABD](https://atisket.pulsewidth.org.uk/?cached=00050087497163-d_268337782-s_3qBn8Lw72Fs6zWyA7yFrMn-i_1590946281)).

kellnerd:
Spotify images no longer have a file extension, so we should not assume that everything after the last dot is a file extension.

Fixes #216.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/seeding/atisket.tsx
+++ b/src/mb_enhanced_cover_art_uploads/seeding/atisket.tsx
@@ -54,7 +54,9 @@ function addSeedLinkToCovers(mbid: string, origin: string): void {
 async function addSeedLinkToCover(fig: Element, mbid: string, origin: string): Promise<void> {
     const imageUrl = qs<HTMLAnchorElement>('a.icon', fig).href;
 
-    const ext = imageUrl.split('.').at(-1);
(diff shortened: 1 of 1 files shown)


ROpdebee:
They actually never had an extension: https://beta.musicbrainz.org/edit/82938734
Oversight on my part, should've tested that change...

ROpdebee:
Not sure what to do with the missing file type though. On the one hand, not having it there might confuse people into thinking it's a bug. On the other hand, it's never been there prior to that commit, and nobody complained 🤷‍♂️

ROpdebee (approved):

ROpdebee:
CAA Dims just requests IA metadata and uses that to extract the mime type they store, so it wouldn't help here in its current state. But I might make some improvements there eventually, e.g. it also doesn't work properly for RG images.

PR #231, opened 2021-11-11
build(dev-deps): replace `node-sass` by `sass`

ROpdebee:
`node-sass` is deprecated and one of its dependencies has a vulnerability. We can use `sass` instead as a drop-in replacement.


ROpdebee:
/deploy-preview to double-check that this doesn't cause a change in the built scripts

kellnerd (approved):
Still working for me 🚢

> We should just be able to use sass as a drop-in replacement for node-sass and rollup-plugin-postcss

Did you mean to say "replacement for node-sass **together with** rollup-plugin-postcss" here? (`rollup-plugin-postcss` is still there and necessary.)

ROpdebee:
Yep, I did, thanks!

PR #234, opened 2021-11-13
fix(caa upload): grab non-public VGMdb images if user is logged in

ROpdebee:
So I totally forgot that `GM.xmlHttpRequest`/`GM_xmlhttpRequest` send cookies in all major engines (of course it does, otherwise, how could you submit edits through userscripts? 🤦 ). This means that we can just use GM xhr to load VGMdb pages and if the user is logged in, we'll be able to extract all images, even the ones that require a login.

We're keeping the VGMdb seeder for now as it might come in handy for some, and it still works, so there's no reason to remove it just yet.

Related to #62 and #197.

Tested with TM/VM/GM4/GM3


src/mb_enhanced_cover_art_uploads/providers/vgmdb.ts
```diff
@@ -146,6 +147,47 @@ export class VGMdbProvider extends CoverArtProvider {
...
+        // istanbul ignore else: Tests are not logged in
+        if (qsMaybe('#navmember', pageDom) === null) {
+            LOGGER.warn('Heads up! VGMdb requires you to be logged in to view all images. Some images may have been missed. If you have an account, please log in to VGMdb to fetch all images.');
+        }
+
+        const coverGallery = qsMaybe('#cover_gallery', pageDom);
+        const galleryCovers = coverGallery ? await VGMdbProvider.extractCoversFromDOMGallery(coverGallery) : [];
+
+        // Add the main cover if it's not in the gallery
+        const mainCoverUrl = qsMaybe<HTMLDivElement>('#coverart', pageDom)?.style.backgroundImage.match(/url\(["']*(.+?)["']*\)/)?.[1];
```
kellnerd, line 166:
```suggestion
        const mainCoverUrl = qsMaybe<HTMLDivElement>('#coverart', pageDom)?.style.backgroundImage.match(/url\(["']?(.+?)["']?\)/)?.[1];
```
Is it possible that the URL is wrapped in multiple quotes?

kellnerd (approved):

src/mb_enhanced_cover_art_uploads/providers/vgmdb.ts
ROpdebee: Probably not 🙂 In JSDom, they're not wrapped in quotes at all, but in FF it appears they are.

PR #235, opened 2021-11-14
feat(caa upload): add RateYourMusic provider

### LINKED ISSUE #191: New provider: RateYourMusic
RateYourMusic (requires login to get link for fullres)

    https://rateyourmusic.com/release/album/fishmans/long-season/ → https://rateyourmusic.com/release/album/fishmans/long-season/buy/ → https://e.snmc.io/i/fullres/w/dd6dc758bde2ed6dfeb5db2b486d97c1/7461038

https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/43?u=ropdebee

Marking as help-wanted because RYM has IP-blocked my VPN.

ROpdebee:
Closes #191 

RYM requires being logged in to get the full-res covers, and thumbnails are WebP so we can't fall back on those if the user isn't logged in. GM.xhr sends cookies, so if the user is logged in, we can get the HTML that contains the URL to the full-res image. We're not logged in in the tests, so I manually edited the recorded response to include the necessary elements for the test (not really ideal, we'll see how much of a burden that becomes when regenerating the recordings). For future reference: Don't replace the whole HTML source, just replace the HTML class name in the profile header and the contents of the `qq` div (which contains the full-res image), since RYM embeds some tokens on the page which seem tied to your account, and it's probably not a great idea to put those in a git repo :smile:


kellnerd (approved):
Since I don't have an account, I couldn't test, so I only had a look at the code.

It's a bit unfortunate that it requires a manual change to update the polly recordings, maybe we can automate that by creating a "patch file" which can be re-applied after the recordings have been updated. Probably it's not as simple as that since the whole HTML is a one-liner in the `.har` files.

PR #246, opened 2021-11-17
chore(coll work attrs)!: remove now-unnecessary userscript

### LINKED ISSUE #242: Deprecate/remove collapse work attrs script
[MBS-11537](https://tickets.metabrainz.org/browse/MBS-11537) got addressed, so I don't think the userscript is necessary anymore.
We should double check that all of the collapsing that happens in the script is also being done on MB. If there's any collapsing which isn't being done, we should keep the script but remove any non-essential collapsing (and probably open a new ticket).

ROpdebee:
MB now collapses work attributes itself: [MBS-11537](https://tickets.metabrainz.org/browse/MBS-11537). They're collapsed on all pages on which we collapse work attributes (works themselves, work tables in artist pages and collections, and edit pages). Therefore, this userscript is now unnecessary and can be removed.

Closes #242


kellnerd (commented):
Maybe we should ship a last update first which notifies the users that the script is no longer necessary?
But I am not sure if a nagging `alert()` message (as it has been done [here](https://github.com/jesus2099/konami-command/commit/c0a8d582da74c9e428ec8316b210a7c9b3a12cbc)) would be appropriate or if we should just blank the script and remove it. Then VM would show an error message once it fails to update the userscript, but I haven't checked what GM and TM do in that case.

ROpdebee:
Hmm that might be a good idea. Perhaps adding a banner to the page (the same as MB's own banner for e.g. beta updates) would be a compromise? I wouldn't go as far as implementing a "dismiss" button because the user should just uninstall the script.

kellnerd:
That's a very good and simple idea, it would only require injecting `<div class="banner warning-header"><p>Please uninstall X...</p></div>` before `#page` (I just tested it with the dev tools).

ROpdebee:
See #253 for deprecation. Let's keep this one open for another few days to allow auto-updates to the deprecated version to take place.

PR #247, opened 2021-11-17
fix(caa upload): improve error message on invalid file types

### LINKED ISSUE #238: Improve error message on unsupported file formats
If we could, we should add the received MIME type in the error message (perhaps extract it from the fetch response header). We should also raise a specific error if we get text or HTML instead of images, since the user may have accidentally pasted the wrong link.

ROpdebee:
* Include actual MIME type returned by server in error message.
* Provide specific error message when content type is textual, as this may mean that the user is trying to use an unsupported provider.

Closes #238

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/fetch.ts
+++ b/src/mb_enhanced_cover_art_uploads/fetch.ts
@@ -207,7 +207,12 @@ export class ImageFetcher {
                 } else if (uint32view[0] === 0x46445025) {
                     resolve('application/pdf');
                 } else {
-                    reject(new Error(`${fileName} has an unsupported file type`));
+                    const actualMimeType = resp.responseHeaders.match(/content-type:\s*([^;\s]+)/i)?.[1];
+                    if (actualMimeType?.startsWith('text/')) {
+                        reject(new Error(`Expected ${fileName} to be an image, but received text. Perhaps this provider is not supported yet?`));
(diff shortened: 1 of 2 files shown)


src/mb_enhanced_cover_art_uploads/fetch.ts
```diff
@@ -207,7 +207,12 @@ export class ImageFetcher {
                 } else if (uint32view[0] === 0x46445025) {
                     resolve('application/pdf');
                 } else {
-                    reject(new Error(`${fileName} has an unsupported file type`));
+                    const actualMimeType = resp.responseHeaders.match(/content-type:\s*([^;\s]+)/i)?.[1];
+                    if (actualMimeType?.startsWith('text/')) {
+                        reject(new Error(`Expected ${fileName} to be an image, but received text. Perhaps this provider is not supported yet?`));
```
kellnerd, line 212: Maybe we should not use the filename here since it is most likely not a filename in that case?
Since the default filename is `image` that often leads to messages like `Failed to grab images: Expected image to be an image, but received text. Perhaps this provider is not supported yet?`. Probably it is already enough to put the "filename" into quotes.

kellnerd (approved):

kellnerd (approved):
Thanks, that's even better than the possible message changes I had in mind.

PR #253, opened 2021-11-19
chore(coll work attrs)!: deprecate now-unnecessary userscript

ROpdebee:
MB offers all of its functionality itself now. We'll push a final update which displays a message to the user, recommending them to uninstall the script. In a few days, we should remove the whole script.

See #246.


kellnerd (approved):
:shipit:

PR #254, opened 2021-11-19
test: E2E tests

### LINKED ISSUE #75: Testing URL match patterns
On a sidenote: Something I would really like to have tests for are the **match patterns** in the userscript's metadata block, since the very unspecific `*` wildcard is rather unsatisfying in most cases and Tampermonkey interprets these patterns differntly from Violentmonkey as we have already noticed.
What do you think about adding test cases for URLs that should or should not be matched to each userscript? As time passes I tend to forget why I designed these patterns in a certain way and it's too easy to miss something during a refactoring or when the scope of the script is extended. A possible solution could be realized by keeping two lists with (representative) "positive" and "negative" URLs inside the `meta.ts` module of each script.
Would it be feasible to create unit tests which would do some basic checks for the match patterns against these lists or is this rather something you would combine with E2E tests because they are userscript engine dependent?

_Originally posted by @kellnerd in https://github.com/ROpdebee/mb-userscripts/pull/68#discussion_r720848981_

ROpdebee:
WIP, experiments. Will update once finalised.

Currently using CodeceptJS for the E2E tests and Selenoid to run different browser versions in Docker containers.
Using CodeceptJS over TestCafe, Puppeteer or Playwright because it can use webdriver (Selenium). E.g. TestCafe uses a proxy so we can't test userscripts: They'll never match the URL. Also chose CodeceptJS over e.g. WebdriverIO because it wraps WebdriverIO (so we can still use it) but for the standard tasks, the interface is much nicer.

We're currently running through Violentmonkey, Tampermonkey, Greasemonkey 4 and Greasemonkey 3. We install userscripts through a temporary localhost server which serves them.

We're currently testing on Firefox latest and 52.0 (lowest version officially supported by MB), and Chrome latest and 64.0 (lowest version supported by MB is 49.0, but it doesn't want to load the extensions for some reason and I can't figure out why. Probably because Violentmonkey supports 55.0 and higher, but it still doesn't work on 55.0, so I selected 64.0 since that's the lowest supported Chrome by Tampermonkey latest. Violentmonkey works on it, so I'm satisfied).

Running locally is a bit tricky, so it's best to look at CI and imitate that.

We're also dropping support for IE and old Safari version here to reduce transpilation (`class` syntax is kept now, for example). Supporting those browsers leads to at least 10% more code in the final bundle, while probably benefiting nobody. Also various optimisations to reduce the size of the built scripts (from 228KB to 152KB currently).

Also adding more polyfills since the E2E tests on the old browsers uncovered incompatibilities. However, we're loading these from [polyfill.io](https://polyfill.io), so for the majority of users, no polyfills will be loaded at all. There's a (possibly big) problem with it though: If multiple userscripts load the polyfill, they might conflict with one another if both polyfills try to define the same property. This happened already while testing URL match patterns, as we have two scripts installed which load the polyfill and both try to define `Symbol.asyncIterator`, which clashed. The current workaround is just to exclude the polyfill for that property altogether, but it's not a robust solution. Need to put more thought into it.

Closes #75 as we've added a test case for URL match patterns too.

TODO:
- [x] Figure out how to install userscript managers
- [x] Install Tampermonkey, Greasemonkey v4 and Greasemonkey v3 (depending on FF version).
- [x] Figure out how to install the userscript(s) in the browser
- [x] Support installing multiple userscripts, build the userscripts beforehand in CI.
- [ ] Write cool E2E tests
- [x] Perhaps look into self-hosting an MB instance instead of using test.MB, as test.MB can be slow at times. Ideally it'd be spun up in CI and thus also destroyed once CI is finished, and wouldn't be exposed to the outside world. **Update:** test.MB's instability makes tests rather flaky, so we should do this if we can.
- [x] Look into uploading E2E test artifacts (videos, screenshots, and logs) as CI artifacts.
- [ ] Can we do coverage analysis in these E2E tests? **Update:** We can.
- [ ] We might need to add some sort of HTTP record-and-replay proxy, or figure out how to integrate PollyJS in the built scripts, to prevent flaky tests.
- [ ] Change CI so that it only runs a couple of the E2E tests on PRs (latest Chrome and Firefox, all major userscript engines) but runs all E2E tests before a deployment. Perhaps add a comment slash command like `/run-full-e2e` to run all E2E tests on request. Due to possible flakiness, perhaps we shouldn't abort deployment if one of the E2E tests on `main` fails, but instead raise an issue for further investigation. If we only test modern browsers on PRs and those E2E tests pass, then it's not _that_ big of a deal if we deploy a version which (temporarily) breaks compatibility with an old browser. If PR E2E tests pass but `main` E2E tests fail, it's probably more likely that that failure is due to flakiness instead of breakages in older browsers.
- [ ] Figure out what to do with polyfill conflicts.

**IMPORTANT**: Make sure to undo 80d9207a58ab7333ac7d108a4dbda65ac3ff89b6 prior to merging.


ROpdebee:
Chrome 64 and FF 52 are failing on these simple E2E tests because of missing polyfills 🙂

ROpdebee:
So I got the E2E setup working with a locally hosted MB instance, but getting it to work in CI is going to be a different story. The docker images take 10 minutes to build, so building them on every run (in the worst case, for every browser/manager combination separately) is not really feasible. I've got an idea for a workaround though, which would pretty much be building the images in a separate repo and pushing them to GitHub's container registry so we can just pull them instead.

Performance of the MB server on a CI machine is a bit of a concern though. Locally, it isn't much faster than going through the real test.MB (and slower than prod MB, but that's to be expected, the local setup isn't tweaked). It might get a bit better if we cull the test DB instance to only contain data that's really necessary, but I wouldn't bet on it being a huge improvement. We don't need the best performance though, we just need increased stability, and test.MB is running an old MBS version too, ideally the tests should be using the latest version. Nonetheless, if the self-hosted MBS eats too many resources of the CI machine, the browser is going to suffer and more steps will time out...

There's a reason why MB's selenium tests are running on a self-hosted jenkins machine 😅  In the worst case, I can probably spin up a self-hosted runner on my VPS though, which should give it a bit more breathing room.

ROpdebee:
Well, this has become a bit of a mess. Merging this PR as-is would lead to a very convoluted history and there are numerous commits whose content has subsequently been reverted. More problematically, this branch has gotten so out of sync with the main branch that resolving the conflicts every couple of days becomes quite a headache. The dependency conflicts, although a pain to deal with, are manageable, but with the refactoring of unit tests + unrelated changes in #299, fixing these conflicts will be difficult.

I'll keep the PR open for now, but I'll start to cherry-pick some groups of changes to smaller PRs so they can already be merged separately, which should hopefully be able to reduce the maintenance overhead of this branch.

ROpdebee:
Why exactly were we using CodeceptJS again? And why are we using a very outdated version of webdriverio? I'm a bit confused…

I may not have read the WDIO docs properly, their test framework seems pretty decent to use… Sure, CodeceptJS is more declarative, but it's also caused plenty of headaches already… I think my main issue with WDIO is that its `describe` and `it` declarations may have conflicted with jest's.

ROpdebee:
I'm getting burnt out by these E2E tests. We'll use these changes for inspiration later, but closing for cleanup.

PR #274, opened 2021-12-05
fix(caa upload): Adapt to new Amazon page source code indentation

### LINKED ISSUE #272: MB: Enhanced Cover Art Uploads Amazon problem
The last few releases I've worked on no longer blow up any Amazon images to full size and is actually limited all of them to 500x?, below what they actually even show on the Amazon page. I can manually open all the images on the Amazon page and Image Max URL blows them up, so I can feed them per link, but when I click on the import from Amazon is when they don't show up larger than 500x? (keeps ratio, but limits to 500)

kellnerd:
The Amazon provider fails to extract the embedded JSON because it is no longer at the beginning of a new line but indented since a recent update.

Fixes #272.


ROpdebee:
Thanks!

PR #299, opened 2021-12-11
test: persist HTTP recordings as WARC files

ROpdebee:
* Clean up PollyJS utilities
* Remove rewiring in tests as it only leads to headaches
* Switch from HAR to WARC format in HTTP recording persistence (see below for motivation)
* Automate "logging in" to RYM during tests (by patching the response automatically rather than manually every time the recordings are generated).

## HAR -> WARC: Motivation
The HAR format is based on JSON, which has numerous problems for our purposes:
1. It's unpleasant for humans to read due to it being JSON.
2. Response content is all collapsed into one line, which makes it infinitely more difficult for humans to read the recording. It's also non-ideal for commits, as the tiniest change will lead to the whole line being changed.

Those two points taken together make it very difficult to "audit" the changes that were made to a recording. Since we re-record these every month, it's quite a chore to check the changes. If we ever decide to automate the re-recording, we need to make it easier to check whether the changes are reasonable.

The WARC format (also standardised, used by e.g. the IA) is plaintext and line-based. It's pretty much serialisation of the actual HTTP messages to a file. WARC files just look a lot better.

On the technical side, we're now using a custom persister which converts from PollyJS' HAR format to a WARC and vice versa, and writes the WARCs to the disk.

Other minor change: We're now using `<suiteName>/<testName>.warc` as the recording path instead of `<suiteName>/<testName>/recording.har` since IMO it makes no sense to put each file in a separate subdirectory and requiring another step to navigate to them.

There are also minor changes to the ECAU source to be able to get rid of `rewire-ts` in the unit tests. None of these should be functional changes though, merely minor refactoring or exporting stuff specifically for tests, so I'm excluding this PR from deployment.


kellnerd:
Good idea, I also considered the HAR format unsuitable for our use case (but I haven't heard about WARC before) since in 99% of the cases we were only interested in the content of the response which had been squeezed into one line.
I had a brief look and the changes looked sensible to me, but I don't have the time to test this further.

PR #323, opened 2021-12-19
Improved Audible Audiobook support

### LINKED ISSUE #322: Improved Audible Audiobook support
Amazon's `/dp/` pages for audiobooks from Audible only have 500px images embedded.
We can get bigger images (usually 2400px) by extracting from the corresponding `hz/audible/mlp/mfpdp/` page for these products.

Example from [Edit #85575314](https://musicbrainz.org/edit/85575314):

* https://www.amazon.co.uk/dp/B07PPB49FK -> https://m.media-amazon.com/images/I/51laMWkzHjL.jpg
* https://www.amazon.co.uk/hz/audible/mlp/mfpdp/B07PPB49FK -> https://m.media-amazon.com/images/I/91gkWujzGwL.jpg (once we support these pages)

kellnerd:
* Support Amazon Audible `/hz/audible/mlp/mfpdp` URLs:
These should have the cover image in its full resolution while Amazon's standard product pages (i.e. `/dp` URLs) provide only 500px images for Audible audiobooks.

* Prefer `/hz/audible/mlp/mfpdp` pages for Audible products:
Look for Audible buttons on standard product pages, as we can only extract 500px images from these pages. In that case we immediately redirect the Amazon provider to `/hz/audible/mlp/mfpdp` pages which should have the same image in its full resolution.

Closes #322 (which is also one of the examples that I have used for testing).

I still remember that you had said that we should probably split Amazon audiobooks into a separate provider in case we would find another special case, which has happended now. Although I had considered this idea, I have not found a clean solution to do this as music and audiobook pages can't be told apart by their URLs.
It probably would not have a been a problem to have two providers for the same URL pattern before this change. But now we have to detect Audible product pages _before_ we even try to extract an image because otherwise the provider would happily return the smaller image for Audible audiobooks. (All the other audiobook logic still can be run _after_ all extraction attempts for a music release have failed.)
If you really want to split the provider, it would require having the audiobook provider run before the music provider. This would introduce a delay of an additional request for the common case of a music release if we don't want to introduce a cache for provider requests.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
+++ b/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
@@ -45,6 +45,14 @@ export class AmazonProvider extends CoverArtProvider {
         const pageContent = await this.fetchPage(url);
         const pageDom = parseDOM(pageContent, url.href);
 
+        // Look for Audible buttons on standard product pages, as we can only
+        // extract 500px images from these pages. Prefer /hz/audible/mlp/mfpdp
+        // pages which should have the same image in its full resolution.
+        if (qsMaybe('.audible_mm_title', pageDom)) {
+            const audibleUrl = new URL(url.pathname.replace(/gp\/product|dp/, 'hz/audible/mlp/mfpdp'), url);
+            return this.findImages(audibleUrl);
+        }
+
         // Look for products which only have a single image, the front cover.
         const frontCover = this.#extractFrontCover(pageDom);
         if (frontCover) {
--- a/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
+++ b/src/mb_enhanced_cover_art_uploads/providers/amazon.ts
@@ -39,7 +39,7 @@ export class AmazonProvider extends CoverArtProvider {
     }
 
     name = 'Amazon';
-    urlRegex = /\/(?:gp\/product|dp)\/([A-Za-z0-9]{10})(?:\/|$)/;
+    urlRegex = /\/(?:gp\/product|dp|hz\/audible\/mlp\/mfpdp)\/([A-Za-z0-9]{10})(?:\/|$)/;
 
     async findImages(url: URL): Promise<CoverArt[]> {
         const pageContent = await this.fetchPage(url);
@@ -73,7 +73,8 @@ export class AmazonProvider extends CoverArtProvider {
     #extractFrontCover(pageDom: Document): CoverArt | undefined {
         const frontCoverSelectors = [
             '#digitalMusicProductImage_feature_div > img', // Streaming/MP3 products
-            'img#main-image', // Audible products
+            'img#main-image', // Audible products, /dp URLs
+            '#mf_pdp_hero_widget_book_img img', // Audible products, /hz/audible/mlp/mfpdp URLs
         ];
 
         for (const selector of frontCoverSelectors) {


src/mb_enhanced_cover_art_uploads/providers/amazon.ts
```diff
@@ -73,7 +81,8 @@ export class AmazonProvider extends CoverArtProvider {
     #extractFrontCover(pageDom: Document): CoverArt | undefined {
         const frontCoverSelectors = [
             '#digitalMusicProductImage_feature_div > img', // Streaming/MP3 products
-            'img#main-image', // Audible products
+            'img#main-image', // Audible products, /dp URLs
```
kellnerd, line 84: Hmm, probably it's possible after all to split this into two providers, if we simply remove this selector for Audible products on `dp` URLs (which would make the music provider fail for audiobooks). The selector should even be no longer in use as the redirect to the other Audible URL format happens before we reach this method (unless I have missed something).
```suggestion
```

kellnerd (commented):
Please let me know what you think, I will update the broken Audible test case once I know whether that's still necessary or if we are going to do the split and will drop this test case.

ROpdebee:
Hmm I'm not really in favour of having two separate Amazon providers with the same URL pattern being present in the dispatch map, that just seems confusing 😄 Is there another way we could reasonably split up the implementation to (clearly) separate it into digital/physical/audiobook? Preferably only by changing the Amazon provider. Simple factory came to my mind, but I don't think that'd be applicable as that would require changes to the dispatcher, IIRC.

(I'm sorry that I have to ask this question, I should know, it's my repository…but I've been doing these damn E2E tests for so long that I forgot what most of the other code looks like 😅 )

kellnerd:
> If you're cool with it, I can go tinker in this branch a bit to see if that'll work.

Sure, let's try that, just push your ideas to this branch, I won't do any further changes until you have checked that approach.

ROpdebee:
Done. I've also removed the thumbnail extraction fallback as it would otherwise silently succeed with inferior images (like #272), it's better to just fail completely so we can fix the extractor for everyone.

I've also locally regenerated all of the recordings and the tests still pass.

src/mb_enhanced_cover_art_uploads/providers/amazon.ts
```diff
@@ -26,6 +27,15 @@ const VARIANT_TYPE_MAPPING: Record<string, ArtworkTypeIDs | undefined> = {
     // See https://sellercentral.amazon.com/gp/help/external/JV4FNMT7563SF5F for further details
 };
 
+// CSS queries to figure out which type of page we're on
+const AUDIBLE_PAGE_QUERY = '.audible_mm_title';
```
kellnerd, line 31: Maybe we should keep my previous comment about the origin of this selector for this line? It would be good to know that this referred to the audiobook button with the Audible logo if Amazon decides to change this (not very meaningful) class name in the future.

kellnerd (commented):
Thanks, I forgot to commit the updated WARC file yesterday.
Looks good to me, there's only one minor thing I would improve before merging. Or is there anything else you would like to change?

ROpdebee:
I've noticed that the placeholder cover detection only works on amazon.com, other domains use images in the local language, and it has a different filename. It's quite easy to fix, but I'm not sure whether it's worth it at this point. We can either:
* Fix it in this PR
* Open an issue and fix it later
* Ignore it until someone complains about it

You decide 🙂

ROpdebee (approved):

kellnerd:
> Open an issue and fix it later

Let's do it the half-lazy way 😇

PR #326, opened 2021-12-21
build: Typescript overhaul

ROpdebee:
Something I complained of some time ago were the implicitly global declarations of libraries like jest, jQuery, etc. This leads to a couple of issues, like jest's `describe`, `expect`, and `it` being "available" in script source code. More recently, this is also causing issues in the E2E tests, where jest's declarations for `describe` etc conflict with WebdriverIO's.

As a solution, I split up the TS project into multiple, smaller subprojects. We now have 5 projects: `build`, `src/lib`, `src/mb_enhanced_cover_art_uploads` (+ one for each other script), `tests/unit`, and a global one that "imports" all of those. There are also some smaller reusable base projects which get extended by the concrete projects, these exist to share configurations. Finally, there's the `./tsconfig.non-composite.json` project which exists for tools which don't support project references (`tsconfig-paths` being one of them, which we use to map e.g. `@lib` to `src/lib` in `ts-node`).

I've also moved some `.d.ts` files closer to where they're relevant, as they didn't need to be `.d.ts` files at all.

"It works on my machine", but @kellnerd, because I changed stuff like `typeRoots` again, can you check that it still behaves as expected in your VSCode? Also, I'm not entirely convinced this is the best possible project structure, so any feedback would be much appreciated. For example, I'm not entirely convinced of having one project for the unit tests while the tested code is split into 3 separate projects, and including `test-utils` in that project as well.


package.json
```diff
@@ -3,11 +3,11 @@
   "version": "1.0.0",
   "description": "Build package for MB userscripts",
   "scripts": {
-    "build": "NODE_ENV=production ts-node build/build.ts",
-    "build-dev": "ts-node build/build.ts",
+    "build": "NODE_ENV=production ts-node -P tsconfig.non-composite.json -r tsconfig-paths/register build/build.ts",
```
kellnerd, line 6: Can you please add a comment why we need a separate `tsconfig.non-composite.json`, `tsconfig.composite.json` and `tsconfig.json`? I don't understand why we need three different configs here and why almost all configs extend `tsconfig.composite.json` while we use `tsconfig.non-composite.json` for the build...
And what's the purpose of the root `tsconfig.json` now, just "keeping things together"?

kellnerd (commented):
Everything is still working fine for me, I have checked a few source files and I haven't noticed any oddities with unrecognised types caused by these changes.
There's only one minor annoyance with the two `.eslintrc.cjs` files, but it does not seem to have any consequences (and likely has already been like that before this PR), see the comment below.

I have to admit that I have no idea how you should set up the tsconfig files for a bigger project, but what you have done so far looks sensible to me. Reviewing the code commit by commit, I thought that there were some redundant tsconfigs in c4c1add7a5c11d134c8d325dec078d4ccb71fdd3, but you have already removed those in 0fd5a325f9399eb6b3c7f027987eea991f201d94. There are still too many of these for my taste, but they all seem to be needed.

> For example, I'm not entirely convinced of having one project for the unit tests while the tested code is split into 3 separate projects, and including test-utils in that project as well.

This is no longer relevant after 0fd5a325f9399eb6b3c7f027987eea991f201d94, right?

package.json
ROpdebee: `tsconfig.base.json` is a base configuration which sets up the main compiler options.
`tsconfig.composite.json` is a base configuration extending `tsconfig.base.json` to set up each concrete project as a composite thing.
`tsconfig.non-composite.json` is a glue project used for tools that don't properly support composite projects, like ESLint and `tsconfig-paths` (`ts-node` is fine I think, but it uses `tsconfig-paths`, so it also needs to use the non-composite glue project). It's equivalent to the original `tsconfig.json`.
`tsconfig.json` is the glue project used by the type checker, which references all composite, concrete projects (through some layers of indirection though).

AFAIK this is like the setup used by TypeScript and `typescript-eslint` themselves. The nomenclature is a bit iffy though, and I don't like how there's like 10 `tsconfig*.json` files scattered around the repository.

Would it perhaps be better to have a directory `configs` that aggregates all of the "base" configurations and actually gives them decent names? And then only have two root `tsconfig.json` files, one for `tsc` and one for `eslint`/`ts-node` etc? If so, should we also put the leaf `tsconfig.json`s in there, or should we keep those in the specific project?

If only `tsconfig.json` had some way to specify which types were available in which directories, then we wouldn't need this mess at all 🙄

.eslintrc.cjs
```diff
@@ -18,7 +18,7 @@ module.exports = {
     parserOptions: {
         ecmaVersion: 12,
         sourceType: 'module',
-        project: './tsconfig.json',
+        project: 'tsconfig.non-composite.json',
```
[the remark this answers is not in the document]
ROpdebee: ESLint is so eager to lint its own configuration that it doesn't stop linting its own configuration :joy: Just enabled it, wasn't done with linting after a minute while it took mere seconds before including it, so I just added it to `.eslintignore`. But now we're typechecking the ESLint configuration though, so..yay?

package.json
ROpdebee: See newer commits.

ROpdebee:
I've consolidated the non-leaf projects into `configs`, containing both the base configurations and the two glue projects (with a couple of comments and a README because I'll definitely forget 90% of the intricacies of a multi-project TS setup). Also simplified a bunch of the leaf projects by moving more options to the base configs. "Fixed" the `eslintrc.cjs` issue by ignoring it (see comment above), but we're now typechecking the tool config files too, and also linting the other configs.

The composite project (now the only `tsconfig.json` in the root) is used for everything except ESLint. ESLint uses the non-composite one (similar to the original project) because otherwise `@typescript-eslint` needs to run TypeScript for each referenced project multiple times, which is slow and can consume too much memory at times.

I'd say it's better than what it was before, but it still isn't great. At least it hides most of the ugly parts in a separate directory 😅

I should perhaps squash some of the commits in this PR prior to merging to prevent the git history having `tsconfig.json` files that only existed for a day.

ROpdebee:
All of the projects seemed to be working properly while I did #340. I did make one change in that PR to `tsconfig.glue-eslint.json` file to also load the type declarations for `jest-extended`, but it doesn't lead to any issues in this specific PR.

kellnerd (commented):
The documentation of the tsconfig files looks good and it is still working for me after I [enabled git symlink support for Windows](https://stackoverflow.com/a/59761201).

.eslintignore
```diff
@@ -3,3 +3,8 @@ coverage/*
 # Ignore top-level scripts for now
 mb_*.js
 lib/*
+node_modules/**
+# ESLint gets really slow when it needs to lint its own configuration. Not sure
+# whether it ever even halts.
+.eslintrc.cjs
+src/.eslintrc.cjs
```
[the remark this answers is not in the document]
ROpdebee: These indeed need a cleanup

ROpdebee:
Hmm, if the symlinks cause a problem, we should perhaps search for a different solution. However, there's also a symlink in the `supported_providers.md` doc, but that then didn't cause issues?

So having symlinks doesn't cause issues, but trying to access the source file does? In that case I could switch the direction of the symlink (put the original in the root and the symlink in the `configs` directory). Or get rid of the symlink altogether and just put that tsconfig in the root. Having the original in the root makes more sense anyway, as the paths in that file are all relative to the root anyway.

kellnerd:
> So having symlinks doesn't cause issues, but trying to access the source file does?

Yes, without having the `core.symlinks` option enabled in git, it checks out a plain text file (which contains the relative file path to the target) instead of a real symlink. I think this setting is/was disabled in Git for Windows by default, as symlinks are a relatively new Windows feature and the underlying command to create them requires special priviledges.

> However, there's also a symlink in the supported_providers.md doc, but that then didn't cause issues?

It did create the text file, but in that case I did not care that there was no working symlink as there was no automatic tool involved, which failed to read its configuration file 😁

> put the original in the root and the symlink in the configs directory

That would be the best solution since we need to a tsconfig.json in there in any case. The symlink in the `configs` directory would be a bonus (as it has no real function other than keeping the configs in the same directory).

tsconfig.json
```diff
@@ -1,48 +1,38 @@
+// Aggregate project referencing all composite sub-projects, for use with `tsc`
+// type checker and `ts-node`.
+// NOTE: Only refer to this file via the symlink in the root, otherwise it will
```
kellnerd, line 3:
```suggestion
// NOTE: Only refer to this file via `tsconfig.json` in the root, otherwise it will
```
The comment is now outdated and should either be fixed or removed.

kellnerd (commented):
LGTM except for the outated comment.

PR #327, opened 2021-12-22
feat(ecau): grab Apple Music original source images

### LINKED ISSUE #80: Apple Music: Fetch original file format
IMU maximises all Apple Music/iTunes covers to a PNG version. However, the source image might not be a PNG. We should add an exception to the maximisation so that we don't unnecessarily consume more bandwidth and storage for a "lossless" version of a lossy source.

For Apple Music, we should be able to determine whether the source image is a JPEG through its URL, see https://beta.musicbrainz.org/edit/81901201. However, old iTunes-style cover art URLs don't contain that filename. The only way to still get those links, AFAIK, is through the old iTunes API. However, a-tisket still uses this old iTunes API, so seeding covers from there will always lead to a PNG unless we skip maximising altogether, which doesn't seem like the right solution. I also doubt we can determine the correct Apple Music URL from an iTunes cover URL. Instead, we should let the a-tisket seeder seed the Apple Music/iTunes release URL which can get processed by the Apple Music provider to determine the Apple Music cover art URL, which then has the original filename through which we can identify the original file format.

All of that is possible, but we should be very careful with _cached_ a-tisket pages. If we seed the release URL itself, the provider will grab the current cover art, even though the cached page may display old cover art which was replaced. So for cached pages, we should fall back to seeding the original cover art URL.

For any direct iTunes cover art links where we cannot or should not use the Apple Music release page to find the Apple Music cover art URL, we should maybe use JPEG images by default instead of PNG.

ROpdebee:
Closes #80 again.

Fast-tracked this because it's been bothering me for quite a while.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/maximise.ts
+++ b/src/mb_enhanced_cover_art_uploads/maximise.ts
@@ -91,27 +91,38 @@ IMU_EXCEPTIONS.set('img.discogs.com', async (smallurl) => {
 // Apple Music
 IMU_EXCEPTIONS.set('*.mzstatic.com', async (smallurl) => {
     // For Apple Music, IMU always returns a PNG, regardless of whether the
-    // original source image was PNG or JPEG. When the original image is a JPEG,
-    // we want to fetch a JPEG version. Although the PNG is of slightly better
-    // quality due to generational loss when a JPEG is re-encoded, the quality
-    // loss is so minor that the additional costs of downloading, uploading,
-    // and storing the PNG are unjustifiable. See #80.
+    // original source image was PNG or JPEG. We can grab the source image
+    // instead. See #80.
 
     const results: MaximisedImage[] = [];
     for await (const imgGeneric of maximiseGeneric(smallurl)) {
         // Assume original file name is penultimate component of pathname, e.g.
         // https://is5-ssl.mzstatic.com/image/thumb/Music124/v4/58/34/98/58349857-55bb-62ae-81d4-4a2726e33528/5060786561909.png/999999999x0w-999.png
-        // We're still conservative though, if it's not a JPEG, we won't
-        // return the JPEG version
-        const pathParts = imgGeneric.url.pathname.split('/');
-        if (/\.jpe?g$/i.test(pathParts[pathParts.length - 2])) {
+        // Transform it to the following:
+        // https://a1.mzstatic.com/us/r1000/063/Music124/v4/58/34/98/58349857-55bb-62ae-81d4-4a2726e33528/5060786561909.png
+        // i.e. change domain to a1, path prefix to us/r1000/063, and drop trailing conversion name.
+        // This should also work for images returned by the old iTunes API.
+        const sourceUrl = new URL(imgGeneric.url);
+        sourceUrl.hostname = 'a1.mzstatic.com';
+
+        // Be a bit defensive and don't try to transform URLs which we don't recognize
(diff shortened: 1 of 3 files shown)


src/mb_enhanced_cover_art_uploads/maximise.ts
```diff
@@ -91,27 +91,38 @@ IMU_EXCEPTIONS.set('img.discogs.com', async (smallurl) => {
...
+        // Transform it to the following:
+        // https://a1.mzstatic.com/us/r1000/063/Music124/v4/58/34/98/58349857-55bb-62ae-81d4-4a2726e33528/5060786561909.png
+        // i.e. change domain to a1, path prefix to us/r1000/063, and drop trailing conversion name.
+        // This should also work for images returned by the old iTunes API.
+        const sourceUrl = new URL(imgGeneric.url);
+        sourceUrl.hostname = 'a1.mzstatic.com';
+
+        // Be a bit defensive and don't try to transform URLs which we don't recognize
+        if (sourceUrl.pathname.startsWith('/image/thumb')) {
+            sourceUrl.pathname = sourceUrl.pathname.replace(/^\/image\/thumb/, '/us/r1000/063');
```
kellnerd, line 110: Although this looks like a country-specific URL, I have just verified that this also works for releases which are not available in the US.

kellnerd (approved):
That was fast! It's great to finally have the original images 🚀

src/mb_enhanced_cover_art_uploads/maximise.ts
ROpdebee: Hmm it's a bit weird. using `nz` seems to return images, but others, like `ca`, `de`, `fr`, `au` etc, lead to an error. I'm not quite sure whether these stand for country codes at all...

ROpdebee: Did some experimentation: You can change the domain name (e.g. to a4.mzstatic.com), still works and returns the same image. You can change the `063` to e.g. `000` or even `99999999999`, still works, identical image. The response headers even barely change. So I'm not sure what these parameters mean exactly.

kellnerd: Ok, let's keep these parameters until we find out more (unless you have the desire to code two lucky numbers into the URL without breaking it).

ROpdebee: It it breaks, we're still emitting the original maximised URL as a backup 🙂

src/mb_enhanced_cover_art_uploads/maximise.ts
```diff
@@ -91,27 +91,38 @@ IMU_EXCEPTIONS.set('img.discogs.com', async (smallurl) => {
...
+        }
+        if (sourceUrl.pathname.split('/').length === 12) {
+            // Drop the trailing conversion filename
+            sourceUrl.pathname = sourceUrl.pathname.split('/').slice(0, -1).join('/');
+        }
+
+        if (sourceUrl.href !== imgGeneric.url.href) {
```
kellnerd, line 117: A little bit late with this observation, but these will always be different since you unconditionally changed `sourceUrl.hostname` in line 106. We should probably compare the `pathname` of these URLs here to prevent duplicate links (probably even broken links because of the domain change).

ROpdebee: Oops. #328.

You weren't late, I was overzealous in merging 🙂

PR #339, opened 2021-12-23
fix(qol): add exception for Washington DC in live comment seeder

### LINKED ISSUE #338: MB: QoL: Live recording comments issue with Washington, DC
https://musicbrainz.org/release/1dbe6954-1d1c-4fe5-a9c2-be42d9baa404

On track 7 when using "live" seeding, for some reason it omitted "Washington". It left "live, 1971‐11‐16: Lisner Auditorium, DC, USA" in the comment. It should be "live, 1971‐11‐16: Lisner Auditorium, Washington, DC, USA"

ROpdebee:
Fixes #338.

### DIFF
--- a/mb_qol_seed_recording_disambiguation.user.js
+++ b/mb_qol_seed_recording_disambiguation.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: QoL: Seed the batch recording comments script
-// @version      2021.12.18
+// @version      2021.12.23
 // @description  Seed the recording comments for the batch recording comments userscripts with live and DJ-mix data.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


mb_qol_seed_recording_disambiguation.user.js
```diff
@@ -129,7 +129,10 @@ function formatRecordingBareArea(area) {
     else country = country.name;
 
     state = state && state.primary_code.split('-')[1];
-    city = city && city.name;
+    // Exception for Washington D.C., it's set as a subdivision in MB, leading
+    // to comments for venues in DC to be "live, ...: <venue>, DC, USA" without
+    // the city name.
+    city = (city && city.name) || (state === 'DC' && 'Washington');
```
ROpdebee, line 135: If there's no city and the state isn't DC, this will be false, meaning it'll be filtered out in the next line. It could be done a bit cleaner though, but this is a "legacy" script which needs to be converted to TS at some point, so we'll clean it up there.

kellnerd (approved):

Issue #351, opened 2021-12-30
Enhanced cover art uploads script fails with "Skipping <filename>: Cannot read properties of undefined (reading 'filename')"


atj:
The latest update to the userscript has completely broken functionality for me, any attempt to import an image results in errors like:

```
Skipping 500x500.jpg: Cannot read properties of undefined (reading 'filename')
```
The error console displays the following:

```
[mb_enhanced_cover_art_uploads] Searching for images in Deezer release…
instrument.js:102 [mb_enhanced_cover_art_uploads] Found 1 images in Deezer release
instrument.js:102 [mb_enhanced_cover_art_uploads] Skipping 500x500.jpg TypeError: Cannot read properties of undefined (reading 'filename')
    at MB： Enhanced Cover A…oads.user.js:106311
    at l (MB： Enhanced Cover A…oads.user.js:104504)
```

The first line in the stack trace points to:

https://github.com/ROpdebee/mb-userscripts/blob/98f99f6c06c7a293b5685df6c23131bf291cb8c9/src/mb_enhanced_cover_art_uploads/fetch.ts#L80

Entering a URL in the input box results in the same error.

I'm running:

- Brave v1.33.106-1 on Linux
- Violentmonkey v2.13.0
- MB: Enhanced Cover Art Uploads v2021.12.29.3

kellnerd:
@atj please install and test the preview from https://github.com/ROpdebee/mb-userscripts/pull/352#issuecomment-1003003221 which should fix the issue.

atj:
Works for me, thanks for the quick fix!

PR #352, opened 2021-12-30
fix(deps): downgrade babel-plugin-transform-async-to-promises to v0.8.16

kellnerd:
Reverts a536aefabbb10500137e68dc335c4ac27b51b521 to hotfix #351.
More notes why this has broken the build at https://github.com/ROpdebee/mb-userscripts/commit/3eacdad5e350196ac64f95627c07cbde342d4053.


ROpdebee:
Did this actually fix it? I'm AFK so I can't test it myself (so thanks @kellnerd for the quick intervention!) but looking at the diff, I don't immediately understand how this upgrade could break anything... it looks like it just enforces a cast to a boolean.

If this didn't fix it, my guess would be the changes made to the async iterators in one of the linting commits. IIRC I added an explicit `return undefined` there, which could explain why `undefined` shows up there.

kellnerd:
It did fix the problem, but I'll admit that I don't fully understand this myself. I think that https://github.com/ROpdebee/mb-userscripts/commit/3eacdad5e350196ac64f95627c07cbde342d4053#diff-bce89bee21387c71137058a0ed02c980600e5ea9ef86bf3fb16538231fff508aR2373 is supposed to return the value of `_iterator.next()` if the `_interrupt` and `_exit` flags are false but v0.8.17 tries to convert it into a Boolean with `!!_iterator.next()` . Also I have not found the root cause in the [babel plugin's changes](https://github.com/rpetrich/babel-plugin-transform-async-to-promises/compare/v0.8.16...v0.8.17) yet.

ROpdebee:
Huh. It seems like there's some conflict between babel's regenerators and the plugin. If we disable regenetators, the plugin transforms the async generator and the `for await` statements itself. But if we enable babel's regenerator, babel transforms the generator and `for await` statements, and then the plugin does some more transformation, and that leads to some conflict...

PR #371, opened 2022-01-02
fix(build): properly transpile async generators

ROpdebee:
Previously, there would be transpilations performed by both babel
itself and the transform-async-to-promises plugin. This led to
conflicts. Instead, we'll let the plugin do all transformations for
async generators, and babel will only do plain generators.

For ECAU, this also completely removes the need for the regenerator
runtime, yet again shrinking the compiled script size significantly.

This is a proper fix for #351 and should allow us to update to 0.8.18 of babel-plugin-transform-async-to-promises.

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -16,6 +16,12 @@ const prodConfig = {
         ['@babel/preset-env', {
             corejs: '3.19',
             useBuiltIns: 'entry',
+            exclude: [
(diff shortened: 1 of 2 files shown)


kellnerd (approved):
I've tested a local build of the userscript, which I produced using version 0.8.18 of the plugin, and it loads images as expected. Thanks for investigating this issue.

PR #374, opened 2022-01-02
chore: re-record HTTP recordings

### LINKED ISSUE #354: Re-record PollyJS HTTP recordings
the tests.

# Why?

# How?

Thank you!

ROpdebee:
I left the two recordings for VGMdb using `vgmdb.info` intact, as `vgmdb.info` is super slow today.

Closes #354.


ROpdebee:
I've been thinking about possibly automating this via GitHub Actions, but then it dawned on me: Is it really worth it committing all of these changes when they don't change the outcome of the tests? We should definitely run the tests without recodings occasionally to find out when upstream changes cause issues in the providers, but there's no point in changing the recordings in this case, as the upstream changes aren't relevant… The only reason I can see to commit these changes is to make sure we don't introduce regressions if we ever change a provider and only test it against an old recording, but there has to be a better way to prevent that. Perhaps CI can do an "integration test" (i.e. one without recordings) only for a specific provider if we make a change to that provider or any code used in that provider…

kellnerd:
> Is it really worth it committing all of these changes when they don't change the outcome of the tests?

Exactly my thought, I would simply ignore the GitHub reminders if the test results do not change due to the updated recordings.

Maybe we should only commit the changes of each specific recording in two cases:
1. The new recording causes one of the unit tests to fail.
2. We change the behaviour of the associated provider (or module, generally speaking).

We could try to do that manually for a while to see if it works well (until we have the desire to also automate that).

ROpdebee:
> We change the behaviour of the associated provider (or module, generally speaking).

100% sure I'll forget to do that, but I think we could do some CI magic to have the bot leave a comment on a PR if the changes affect a file which eventually affects a test file (`jest --listTests --changedSince main`) that uses a recording, without changing the recording. It might produce false positives on style changes and it'll probably complain if we change something in `@lib` that gets used in the providers, but in those cases we can just ignore the warning.

ROpdebee:
[Probot](https://probot.github.io/) could help automate this :)

PR #375, opened 2022-01-02
fix(work codes): support translated agency IDs

### LINKED ISSUE #349: Bulk copy-paste work codes does not work in other MB languages
I have a few MB userscripts installed, but this one does not figure out.  Bulk copy-paste work codes v2021.10.20, greasemonkey, Firefox 95.0.2 (64-Bit). Windows 10 Home. v. 21H1. build 19043.1415, Windows Feature Experience Pack 120.2212.3920.0. 
Button appaers on iswcnet.cisac.org ("Copy work codes"). Paste button appears,too: Fill work codes / Format work codes. But nothing is getting pasted.

ROpdebee:
In some languages, the agency code keys are translated leading to
issues pasting the codes. For example, in German, the key for ASCAP
is "ASCAP-ID" while the script expects the English "ASCAP ID". This
is a temporary fix to normalise such translations. Some exceptions,
most notably PRS, may still not work correctly in translations. We
need a better fix for the script to work in other languages, but for
that, some parts would need quite invasive changes.

Fixes #349.

### DIFF
--- a/mb_bulk_copy_work_codes.user.js
+++ b/mb_bulk_copy_work_codes.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Bulk copy-paste work codes
-// @version      2021.10.20
+// @version      2022.1.2
 // @description  Copy work identifiers from various online repertoires and paste them into MB works with ease.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


kellnerd (approved):
Good enough as a temporary fix.

> For example, in German, the key for ASCAP is "ASCAP-ID" while the script expects the English "ASCAP ID".

Sorry, that's "my fault", I have translated these into German a while ago. But apparently I wasn't the only one who was bored enough to translate these strings 😂

PR #376, opened 2022-01-03
build: reduce renovate spam

ROpdebee:
Allow renovate to automatically merge certain dependency updates (patch and minor bumps to dev dependencies; pinning dependencies, and lock file maintenance) without even creating a PR. This should reduce many of the notifications.

Renovate will check that all CI jobs pass before attempting to merge. I've added a new CI job that will fail build/test/lint/typecheck as normal, but will also fail if the built scripts would be changed in any way. This ensures that we don't let Renovate automatically merge updates to dev dependencies that cause changes in production (e.g. babel). In case of failures, Renovate will open a real PR instead.

We'll still get real PRs in the following cases:
* Major bumps to dev dependencies (on weekends)
* Any update to prod dependencies (on weekends)
* Any dev patch/minor which causes a change in the userscripts or otherwise fails CI (could happen during the week)

The reason I didn't go with grouped updates is because of the issues we've already faced with `node-fetch` and because the whole group is one big commit, meaning that if anything goes wrong, we'd have to revert _all_ updates only to re-apply them later on again. Grouping is a good way to reduce spam, but allowing renovate to automerge when it is safe to do so just by merging branches without creating PRs will reduce it even more.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,5 +5,22 @@
     ":semanticCommitTypeAll(build)",
     ":maintainLockFilesMonthly",
     ":label(dependencies)"
+  ],
+  "packageRules": [
+    {
+      "description": "Automatically merge pinning dependencies and lock file maintenance",
+      "matchUpdateTypes": ["pin", "lockFileMaintenance"],
+      "automerge": true,
+      "automergeType": "branch",
+      "schedule": ["at any time"]
+    },
+    {
+      "description": "Automatically merge patch and minor updates to dev dependencies",
+      "matchDepTypes": ["devDependencies"],
+      "matchUpdateTypes": ["patch", "minor"],
+      "automerge": true,
+      "automergeType": "branch",
+      "schedule": ["at any time"]
+    }
   ]
 }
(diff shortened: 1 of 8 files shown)


kellnerd (approved):
I still don't speak GitHub Actions and can't test this, but the suggested changes and the TS code look good to me.

ROpdebee:
Well I'm not sure it'll work, but we'll find out soon enough :)

ROpdebee:
It worked, sort of. Renovate basically immediately created the branches and commits for the two pending updates, without creating a PR. CI ran on both branches and renovate now states that the updates are "awaiting status checks" (which in this case really means that it's awaiting the next run of renovate on this repository). It'll take another few iterations since it'll have to rebase a couple of times, but it _should_ work. The only potential problem could be that pushes to `main` require certain status checks to be completed, but it's running those status checks anyway.

However.. it's running those checks through a different workflow, so it might still fail.

Also, for some reason CI is skipping the check to see whether userscripts are being changed, so that has to be fixed too.

ROpdebee:
It worked! Took some debugging though. But no more spam!

PR #386, opened 2022-01-24
feat(ecau): improve error message on unsupported provider URLs

ROpdebee:
When a user e.g. inputs a Discogs master URL, we should indicate that it's not a valid URL for the Discogs provider, rather than suggest that the provider does not yet exist.


kellnerd (approved):
Good idea, I could also imagine there might be yet another different Amazon page type which is not supported yet 👍

PR #387, opened 2022-01-24
feat(ecau): add DatPiff provider

### LINKED ISSUE #332: ECAU: DatPiff provider
DatPiff (not yet supported by Image Max URL either it seems)

[Edit #84281792 - MusicBrainz](https://beta.musicbrainz.org/edit/84281792)

### LINKED ISSUE #388: ECAU: Improve logging when placeholder images are removed
Cf. @kellnerd's comment in #387

Creating a new issue so I won't forget :innocent:

ROpdebee:
✅ Updated `supported_providers.md`

Nothing too special about this one, mainly HTML parsing, although it does need some postprocessing to filter out dummy placeholders by content digest (the placeholders are binary identical but have different URLs). Also requires custom maximisation logic.

About the postprocessing: I just discovered (while submitting the PR) that the requests actually return an ETag in the response header, which is the MD5 digest of the content. We could use that instead of computing the digest ourselves, but OTOH, that would require somehow providing the HTTP response to the postprocessing method, which we're currently not doing yet.

Closes #332.

> Also closes #388 which I unnecessarily created because I thought the problem of missing log messages was more widespread than it truly is :)


tests/unit/mb_enhanced_cover_art_uploads/providers/datpiff.test.ts
```diff
@@ -0,0 +1,98 @@
...
+                index: 0,
+                urlPart: 'md09671e/B_unique_Stay_Sharp_2-front',
+                types: [ArtworkTypeIDs.Front],
+            }, {
+                index: 1,
+                urlPart: 'md09671e/B_unique_Stay_Sharp_2-back',
+                types: [ArtworkTypeIDs.Back],
+            }],
+        }, {
+            desc: 'mixtape without front cover',
```
kellnerd, line 55:
```suggestion
            desc: 'mixtape without back cover',
```

kellnerd (commented):
Looks good, but I was surprised that the script claimed "Found 1 images in DatPiff release" but queued none for upload when I did my first test with a random URL from the unit tests. Then I realized that I had chosen the release with the placeholder image 😂

I am not sure whether I had ever tested with a placeholder image release from a different provider before, but I think there should be a better final logger message which explains what has happened to the image (instead of the "Successfully added X image(s)" message which of course does not replace the older "Found X images in DatPiff release" message* in this case).

*) By the way, the plural form of image is not wrapped in brackets here 🤓

ROpdebee:
Also closes #388 which I unnecessarily created because I thought the problem of missing log messages was more widespread than it truly is :)

tests/unit/mb_enhanced_cover_art_uploads/providers/datpiff.test.ts
```diff
@@ -52,7 +52,7 @@ describe('datpiff provider', () => {
                 types: [ArtworkTypeIDs.Back],
             }],
         }, {
-            desc: 'mixtape without front cover',
+            desc: 'mixtape without back cover',
```
kellnerd, line 55: One of the ugliest places to change a string, the WARC file also has to be renamed now...

ROpdebee: Oops, totally forgot about that. It looked like such a trivial change

kellnerd (approved):

PR #390, opened 2022-01-25
feat(ecau): archive.org provider

### LINKED ISSUE #275: [Enhanced Cover Art Uploads] Add archive.org as a provider
Great script, could you add support for archive.org? Example: https://musicbrainz.org/release/456117de-63a3-4842-af87-a4f4f20c61b8/ which is linked to https://archive.org/details/20010917Fantmas-MagicStickDetroitMIUSA

Thanks!

ROpdebee:
Closes #275.

Also changes the MB provider to delegate to the IA provider, since the IA provider will eliminate the redirects (and therefore also the useless redirect warnings). + Added a small CAA provider that does pretty much the same thing.

`supported_providers.md` updated.


src/mb_enhanced_cover_art_uploads/providers/archive.ts
```diff
@@ -0,0 +1,133 @@
...
+            throw new Error('Cannot extract images: This item is darked');
+        }
+
+        return itemMetadata;
+    }
+
+    private createBaseDownloadUrl(itemMetadata: ArchiveMetadata): string {
+        // While we could just use the standard archive.org/download/... URL,
+        // it would always lead to redirection warnings which can be avoided.
+        return `https://${itemMetadata.server}${itemMetadata.dir}/`;
```
kellnerd, line 131: It's a shame that ES still has no obvious clean solution to join URL components without external libraries and you often have to resort to string manipulation.
```suggestion
    private createBaseDownloadUrl(itemMetadata: ArchiveMetadata): URL {
        // While we could just use the standard archive.org/download/... URL,
        // it would always lead to redirection warnings which can be avoided.
        return new URL(itemMetadata.dir, `https://${itemMetadata.server}`);
```
... would be a possible solution here (still not pretty because of the protocol), but as soon as you want to append path components to a non-root URL it starts to become ugly, e.g. `new URL('./' + path, baseDownloadUrl)` above. On the other hand ```new URL(`./${itemId}-${imageFileName}`, baseDownloadUrl)``` would look a lot better, so what do you think?

src/mb_enhanced_cover_art_uploads/providers/musicbrainz.ts
```diff
@@ -1,50 +1,28 @@
...
+        // Delegate to Archive.org provider to prevent useless redirects.
+        return new ArchiveProvider().findImagesCAA(`mbid-${mbid}`);
     }
 }
+
+// Mostly identical to MB provider, but a different domain and name.
+export class CoverArtArchiveProvider extends MusicBrainzProvider {
+    override supportedDomains = ['coverartarchive.org'];
+    override favicon = 'https://coverartarchive.org/favicon.png';
+    override name = 'CoverArtArchive';
```
kellnerd, line 26:
```suggestion
    override name = 'Cover Art Archive';
```
See above.

kellnerd (commented):
Code looks good to me, tested with CAA URL, MB URL and IA URL.

src/mb_enhanced_cover_art_uploads/providers/archive.ts
```diff
@@ -0,0 +1,133 @@
...
+        const itemMetadata = safeParseJSON<ArchiveMetadata>(itemMetadataResp, 'Could not parse IA metadata');
+
+        // IA's metadata API always returns a 200, even for items which don't
+        // exist.
+        if (!itemMetadata.server) {
+            throw new Error('Empty IA metadata, item might not exist');
+        }
+
+        if (itemMetadata.is_dark) {
+            throw new Error('Cannot extract images: This item is darked');
```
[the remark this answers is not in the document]
ROpdebee: "darked" is IA terminology, but yes, I do agree.

src/mb_enhanced_cover_art_uploads/providers/archive.ts
ROpdebee: `new URL(path, baseDownloadUrl)` should work fine, as long as `baseDownloadUrl` ends in a slash.

ROpdebee: I went with a small helper function because the URL constructor's argument order is backwards, which makes no sense IMO.

The helper could be further extended to add the slashes where necessary or even take a variable list of arguments to join an arbitrary number of URL parts, but that's not necessary at the moment and would require unit testing, and I'm feeling lazy and don't want write those today 😅

kellnerd (approved):

PR #391, opened 2022-01-25
feat(ecau): seed from marlonob's atisket

### LINKED ISSUE #251: Enhanced Cover Art Uploads on etc.marlonob.info
Could you please add this to https://etc.marlonob.info/atisket/ as well? I use it alot because the mirror removes a lot of Deezer countries.

ROpdebee:
Contrary to atj's mirror, since we can't easily get the release ID from marlonob's original version, we don't seed the release URL but seed the direct image URL instead. We did the release URL workaround for atj's mirror so that we could properly maximise the Apple Music images to JPEG/PNG depending on source format, but we now have a better maximisation strategy for Apple Music, so it's no longer relevant.

Closes #251


kellnerd (approved):
LGTM, but did not test.

PR #393, opened 2022-01-25
feat(ecau): bump IMU to latest version

### LINKED ISSUE #392: Drop IMU exceptions which are no longer necessary
We can probably remove our custom maximisation logic for Apple Music by now, IMU has implemented support for source images in https://github.com/qsniyg/maxurl/commit/56e0fce886bd328e8a155fc69dd62ff954fa0053 (but I have not tested it yet).

Same for 7digital, native support has been added in https://github.com/qsniyg/maxurl/commit/9e50ab7dff78df686b4422dcbcb4cf418ba5df22.

ROpdebee:
More maximisation!

Exceptions for 7digital have been removed, exception for Apple Music has been changed to mark the `/source` URLs that are likely broken as such, and we'll skip warning the user when we can't fetch a likely broken image.

Closes #392.

We should also perhaps create some more IMU tickets to add Jamendo and DatPiff.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/maximise.ts
+++ b/src/mb_enhanced_cover_art_uploads/maximise.ts
@@ -331,21 +331,6 @@ IMU_EXCEPTIONS.set('*.mzstatic.com', async (smallurl) => {
     return results;
 });
 
-// IMU has no definitions for 7digital yet, so adding an exception here as a workaround
-// Upstream issue: https://github.com/qsniyg/maxurl/issues/922
-IMU_EXCEPTIONS.set('artwork-cdn.7static.com', async (smallurl) => {
-    // According to https://docs.7digital.com/reference#image-sizes, 800x800
-    // and 500x500 aren't available on some images, so add 350 as well as a
-    // backup.
-    return ['800', '500', '350'].map((size) => {
-        return {
-            url: new URL(smallurl.href.replace(/_\d+\.jpg$/, `_${size}.jpg`)),
-            filename: '',
-            headers: {},
-        };
-    });
-});
-
 IMU_EXCEPTIONS.set('usercontent.jamendo.com', async (smallurl) => {
     return [{
         url: new URL(smallurl.href.replace(/([&?])width=\d+/, '$1width=0')),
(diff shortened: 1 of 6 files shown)


ROpdebee:
Not sure what to do with the coverage here. It feels stupid to add a couple of test cases just to cover those two harmless statements…

src/mb_enhanced_cover_art_uploads/maximise.ts
```diff
@@ -293,60 +294,30 @@ IMU_EXCEPTIONS.set('img.discogs.com', async (smallurl) => {
...
-            sourceUrl.pathname = sourceUrl.pathname.replace(/^\/image\/thumb/, '/us/r1000/063');
+        if (urlBasename(imgGeneric.url) === 'source') {
+            lastSource = imgGeneric;
+        } else if (
+            typeof lastSource !== 'undefined'
+                && urlDirname(imgGeneric.url) === urlDirname(lastSource.url)
+                && imgGeneric.url.hostname === lastSource.url.hostname) {
+            // Mark the `/source` image as likely broken if there's another URL
+            // with the same directory but a different image name.
+            lastSource.likely_broken = true;
```
kellnerd, line 313: FWIW, https://a1.mzstatic.com/us/r1000/063/Music111/v4/e1/dc/68/e1dc6808-6d55-1e38-a34d-a3807d488859/191061355977.jpg still shows the "Skipping maximised candidate source: HTTP error 404: Not Found" warning as `source` is a 404 (example taken from https://github.com/qsniyg/maxurl/commit/56e0fce886bd328e8a155fc69dd62ff954fa0053). So this is not always working as expected?

ROpdebee: Hmm, for that URL IMU only returns one candidate, so it's never executing the iteration that would mark it as `likely_broken`. Will rewrite the logic to not rely on multiple candidates, it should instead compare the maximised URL to the one that was given as input.

ROpdebee: Should be fixed now.

kellnerd (approved):
Looking good now, please don't forget to update/commit the WARC files.

ROpdebee:
I don't think there are any changes affecting the recordings, since the maximisation tests don't actually perform any requests. However, I cleared the recordings and re-tested anyway, which lead to two unrelated test failures which I'll fix afterwards.

PR #395, opened 2022-01-26
feat(ecau): improved logging message display

### LINKED ISSUE #81: Improve status banner GUI

ROpdebee:
I finally went ahead and did it, the new status message GUI has arrived 🎉
Also improves logging of current progress while fetching.
Fixes #81.

Now that we have a bit more real estate to work with, perhaps it's time to rewrite the log messages to log the full URL instead of just the filename? E.g. Discogs changed their URLs _again_ and now they always have the same filename for the same release.

### DIFF
--- a/src/lib/logging/guiSink.scss
+++ b/src/lib/logging/guiSink.scss
@@ -1,0 +1,40 @@
+#ROpdebee_log_container {
+    margin: 1.5rem auto;
+    width: 75%;
+
+    .msg {
+        padding: .5rem .75rem;
+        margin-bottom: .5rem;
+        width: 100%;
+        display: block;
+        border: 1px solid;
+        border-radius: 4px;
+        font-weight: 500;
+
+        // Colours come from bootstrap.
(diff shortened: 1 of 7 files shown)


kellnerd (approved):
I definitely like the new look, it also looks good with a dark theme.

ROpdebee:
Opinions on changing the log messages to use full URLs?

kellnerd:
> Opinions on changing the log messages to use full URLs?

I think it has advantages for providers which return meaningless filenames or cause errors and it should cause no harm in other cases. There shouldn't be multiple messages with filenames/URLs at the same time, so the chance they could clutter the page is very low, I guess.
In most cases these messages have already disappeared before you can notice a difference, so I hadn't really thought about this yet. But it would probably be useful for potential issue reports.

PR #397, opened 2022-01-26
fix(ecau): new Discogs image URLs

ROpdebee:
* Fix maximisation of single URLs.
* Fix displayed filenames.

I don't like that we have to do this, but it is what it is. Stupid Discogs with their security through obscurity.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/providers/discogs.ts
+++ b/src/mb_enhanced_cover_art_uploads/providers/discogs.ts
@@ -1,6 +1,5 @@
 import { assert, assertHasValue } from '@lib/util/assert';
 import { safeParseJSON } from '@lib/util/json';
-import { urlBasename } from '@lib/util/urls';
 import { gmxhr } from '@lib/util/xhr';
 
 import type { CoverArt } from './base';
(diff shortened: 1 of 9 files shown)


ROpdebee:
TBH we should perhaps just drop the Discogs maximisation exception altogether. The main reason for its existence was to prevent IMU from returning webp images, but it won't be able to anymore, since Discogs doesn't serve those any longer. The Discogs provider returns maximised images anyway, so we'd only need to account for the thumbnails that people might enter manually. However, getting those thumbnails from the Discogs source page is difficult because of their DOM structure (at least in Firefox, it does not give you the option to copy the image source URL from the thumbnail without opening the image viewer, and the image viewer gives the larger version). Secondly, we should just delegate that maximisation to IMU (but we'd need to allow it to perform additional requests to maximise an image, which should be doable), assuming they'll fix this on their end too.

The downside would be the non-descriptive filenames, but again, an upstream IMU patch would likely fix that, and besides, Discogs is not the only provider with bad file names.

src/mb_enhanced_cover_art_uploads/providers/discogs.ts
```diff
@@ -109,17 +108,32 @@ export class DiscogsProvider extends CoverArtProvider {
...
         // Maximising by querying the API for all images of the release, finding
         // the right one, and extracting the "full size" (i.e., 600x600 JPEG) URL.
-        const imageName = url.pathname.match(/discogs-images\/(R-.+)$/)?.[1];
+        const imageName = this.getFilenameFromUrl(url);
         const releaseId = imageName?.match(/^R-(\d+)/)?.[1];
 
         /* istanbul ignore if: Should never happen on valid image */
         if (!releaseId) return url;
         const releaseData = await this.getReleaseImages(releaseId);
+        console.log(releaseData.data.release.images.edges);
```
kellnerd, line 134:
```suggestion
```
An old debugging statement, I guess.

src/mb_enhanced_cover_art_uploads/providers/discogs.ts
```diff
@@ -109,17 +108,32 @@ export class DiscogsProvider extends CoverArtProvider {
         return metadata;
     }
 
+    static getFilenameFromUrl(url: URL): string | undefined {
+        // E.g. https://i.discogs.com/aRe2RbRXu0g4PvRjrPgQKb_YmFWO3Y0CYc098S8Q1go/rs:fit/g:sm/q:90/h:600/w:576/czM6Ly9kaXNjb2dz/LWltYWdlcy9SLTk4/OTI5MTItMTU3OTQ1/NjcwNy0yMzIwLmpw/ZWc.jpeg
```
kellnerd, line 112: I'm curious, did you find any documentation about this URL pattern or did you just try to decode parts of the URL because it looked like Base64?

kellnerd (approved):
Completely agree with what you said, let's hope that IMU offers better support for the new Discogs URLs. Until then this patch is a welcome improvement.

> we should just delegate that maximisation to IMU (but we'd need to allow it to perform additional requests to maximise an image, which should be doable)

I do not fully understand what you mean with "allow[ing] it to perform additional requests" here, isn't this something which IMU is already doing, e.g. for Deezer?

ROpdebee:
> I do not fully understand what you mean with "allow[ing] it to perform additional requests" here, isn't this something which IMU is already doing, e.g. for Deezer?

Huh, yeah. I think I was confusing it with allowing IMU to use 3rd party services for maximisation.

src/mb_enhanced_cover_art_uploads/providers/discogs.ts
ROpdebee: I spotted parts ending in '=', which is a telltale sign of base64, so I messed around with trying to decode it and figured out the file name part. The signature was a case of unsuccessful trial-and-error to see if I could figure anything out about it, also using the old URL format, after which I looked at the docs on Thumbor (which Discogs used to use according to some IMU ticket) and found out the signature part in the old URLs. Then I looked up alternatives to Thumbor to compare with the new format, found imgproxy with the exact same URL structure, and then I called it quits since I highly doubt anyone will be able to reverse-engineer a maximisation strategy without resorting to things they shouldn't do (hacking/brute forcing). Brute forcing would be absolutely useless anyways because as soon as the key is found, they'll just change it.

I'm like 99% sure someone over at Discogs is keeping an eye on all of the threads which try to figure out how to get larger images, with a sort of evil laugh knowing that it's nigh impossible to do so. To that person: 🖕

PR #398, opened 2022-01-26
feat(ecau): add provider buttons to main cover art page

### LINKED ISSUE #384: Add provider import buttons to main cover art page too
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/46?u=ropdebee

ROpdebee:
Closes #384.


kellnerd (approved):
Tested and working for me (although it feels strange to seed the upload page from MB itself and when it's just a click away).

ROpdebee:
> although it feels strange to seed the upload page from MB itself and when it's just a click away

Agreed, but I'm not aware of another straightforward way to send the URL from one page to the next. Having it implemented as a seeder is pretty weird though, but that meant we can reuse the page detection that already exists, rather than having to implement another exception.

PR #401, opened 2022-01-27
perf(ecau): refactor sync generators to prevent babel regenerator

### LINKED ISSUE #399: ECAU 2021.1.26.2 added an unwelcome guest
babel's regenerator-runtime is back in the compiled script since 2021.1.26.2 and that bloats the script size. I guess it's added again because of the `enumerate` generator function, but it's not the only generator function in the project, so I'm not sure why that specific one triggered the import. I'll have to look into it in the near future.

ROpdebee:
Fixes #399.

### DIFF
--- a/src/lib/util/array.ts
+++ b/src/lib/util/array.ts
@@ -41,9 +41,6 @@ export function collatedSort(array: string[]): string[] {
     return array.sort(coll.compare.bind(coll));
 }
 
-export function* enumerate<T>(array: T[]): Iterable<[number, T]> {
-    let idx = 0;
-    for (const el of array) {
-        yield [idx++, el];
-    }
+export function enumerate<T>(array: T[]): Array<[T, number]> {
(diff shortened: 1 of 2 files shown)


kellnerd (approved):
The alternative would be to use `Array.forEach()` which also provides the index to its callback function. That would be doable for our current use case, but it's not always feasible to use a callback function instead of an inline loop.

ROpdebee:
> That would be doable for our current use case

It wouldn't preserve the current semantics, since there's async stuff going on in the loop body. While we could use `Array.map` with `Promise.all`, that would lead to multiple images being fetched in parallel and the log messages showing up out-of-order. That could still be acceptable though.

Speaking of `Array.map`, why am I not just doing `array.map((el, idx) => [el, idx])`? 🤔

kellnerd (approved):
You are right, I had missed the `await` keyword.

PR #403, opened 2022-01-27
fix(ecau): Amazon Audible covers

### LINKED ISSUE #400: Amazon audiobooks causing trouble again
https://www.amazon.fr/gp/product/207055998X (an audiobook on CD) fulfills both the `PHYSICAL_AUDIOBOOK_PAGE_QUERY` and the `AUDIBLE_PAGE_QUERY` (because a digital media version is also available on Audible). Since the `AUDIBLE_PAGE_QUERY` condition comes first, the extraction of the cover images fails: The Audible release has a different ASIN, so the attempt to find the non-existing `/hz/audible/mlp/mfpdp/` page of the CD leads to a redirect to the product page, where the `AUDIBLE_FRONT_IMAGE_QUERY` does not work of course.
Changing the order of the conditions fixes this case but breaks our existing Audible unit test.
I don't have time to look into it right now, but we need better page queries if we don't want to introduce a fallback mechanism again.

ROpdebee:
Prevent the Amazon provider from matching a physical audiobook page which have an Audible variant, as the Audible variant. Also regenerated Amazon recordings to ensure we don't break anything.

Fixes #400.

I've used the .com version in the unit tests because the .fr version caused a 503 and wanted me to fill a captcha (probably missing a cookie or something). The pages of both are the same though, and the .fr page works in the browser too, it's just the unit tests.

Bonus fix: While testing I discovered that Audible product URLs like https://www.amazon.fr/Harry-Potter-%C3%A0-l%C3%89cole-Sorciers/dp/B06Y67NZ77 lead to a 404 because the provider was transforming them to https://www.amazon.fr/Harry-Potter-%C3%A0-l%C3%89cole-Sorciers/hz/audible/mlp/mfpdp/B06Y67NZ77, which isn't correct. That's now been fixed too.


src/lib/util/urls.ts
```diff
@@ -6,6 +8,11 @@ export function urlBasename(url: string | URL, defaultBasename = ''): string {
     return url.split('/').pop() || defaultBasename;
 }
 
-export function urlJoin(base: string | URL, sub: string): URL {
-    return new URL(sub, base);
+export function urlJoin(base: string | URL, ...subPaths: string[]): URL {
+    assert(subPaths.length >= 1);
+    let newUrl = new URL(subPaths[0], base);
+    for (const nextSub of subPaths.slice(1)) {
+        newUrl = new URL(nextSub, newUrl);
```
kellnerd, line 15:
```suggestion
    let newUrl = new URL(base);
    for (const subPath of subPaths) {
        newUrl = new URL(subPath, newUrl);
```
I think that's a bit more elegant 😉

src/lib/util/urls.ts
```diff
@@ -1,3 +1,5 @@
+import { assert } from './assert';
+
```
kellnerd, line 2:
```suggestion
```
No longer necessary, see my next suggestion.

kellnerd (approved):
Thanks a lot. (It seems that I should not report issues if I intend to fix them myself at a later time 😅)

ROpdebee:
> It seems that I should not report issues if I intend to fix them myself at a later time 😅

I just happened to log in at around the same time as you submitted the issue, so it seemed like a good start to the day 😄

PR #404, opened 2022-01-27
chore(ecau): recording cleanups and upstream fixes

ROpdebee:
* Removed unused recordings
* Fix QUB Musique test case as the release we were using to test, was deleted
* Spotify changed their 404 page and is no longer returning a 404 status code, so we now do (simple) custom detection. Also, this is PR number 404. Coincidence?!?! probably

I've also updated the Qobuz docs to note that the maximised images are untouched originals, which I discovered while working on the QUB fixes by looking at the EXIF metadata.


kellnerd (approved):
> Also, this is PR number 404. Coincidence?!?! probably

Exactly my thought when I read the email notification 😂

PR #417, opened 2022-02-28
Create a GEMA repertoire handler

kellnerd:
The GEMA repertoire handler is fairly straightforward (there is only one agency code and one ISWC per work which has to be extracted) and works for the English and the German search. Pagination of the results does not cause problems, the mutation observer correctly catches all these events, except for the inital page load which might load the results of the last visit again and requires an inital handler.

It took me a while to figure out why the code of the first commit alone failed to work, although I've implemented it the same way as the CISAC handler. Then I found out that I have to drop the explicit JSON serialization of the "clipboard data" and directly feed the raw object into the GM storage function to make it work.

Violentmonkey and Tampermonkey already do the JSON serialization implicitly while
Greasemonkey v4 has never supported this userscript (which uses GM v3 API).
This is necessary because https://online.gema.de/werke/search.faces apparently overwrites `JSON.stringify()` with a custom implementation (try `JSON.stringify({a:[]})` in a browser console on their page) which causes `JSON.parse()` to fail on MB.

### DIFF
--- a/mb_bulk_copy_work_codes.user.js
+++ b/mb_bulk_copy_work_codes.user.js
@@ -308,8 +308,7 @@ class BaseWorkForm {
         return this.findEmptyRow(parentSelector, inputName);
     }
 
-    checkAndFill(rawData) {
-        let data = this.parseData(rawData);
+    checkAndFill(data) {
         console.log(data);
         let externalCodes = extractCodes(data);
         let externalISWCs = data['iswcs'];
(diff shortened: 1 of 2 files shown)


mb_bulk_copy_work_codes.user.js
```diff
@@ -699,8 +695,73 @@ function handleISWCNet() {
...
+//////////////
+// GEMA
+//////////////
+
+function handleGEMA() {
+
+    function findAgencyWorkCodes(tr) {
+        return {
+            'GEMA': [
+                tr.querySelector('.workSocworkcde').innerText.match(/(\d{0,8})[\-‐](\d{3})/)[0],
```
kellnerd, line 708: That's copied verbatim from the [validation library](https://github.com/ROpdebee/mb-userscripts/blob/d27189dfd408547ed8f5ab07f08a8fc17e317f10/lib/work_identifiers.js) as it currently doesn't expose the agency patterns.
I don't think it would be worth the effort, given that the pattern is only a fancy way to strip the "GEMA work no.:" or "GEMA-Werk.-Nr:" prefix.
But I'm wondering if the code shouldn't have at least one digit before the version code?
```suggestion
                tr.querySelector('.workSocworkcde').innerText.match(/(\d{1,8})[\-‐](\d{3})/)[0],
```

ROpdebee: Probably yes, not sure why it allows an empty code...

ROpdebee (approved):
LGTM, but haven't tested personally. Sorry for the delay!

ROpdebee:
Ugh, I need to fix that so that it doesn't deploy leftover skipped changes from a previous PR.

PR #429, opened 2022-03-31
Restore `JSON.stringify()` for GEMA repertoire search

kellnerd:
See https://github.com/ROpdebee/mb-userscripts/commit/1a163ace55acb4cd29d04c3e6dae8b6c0ea8dc74#commitcomment-69876519.

A dirty workaround for a dirty hack which is used by the GEMA website. I would have preferred my original solution to simply drop the JSON serialization, but if this helps to make the userscript work again with Chrome+Tampermonkey, let's do that 😝

### DIFF
--- a/mb_bulk_copy_work_codes.user.js
+++ b/mb_bulk_copy_work_codes.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Bulk copy-paste work codes
-// @version      2022.2.28
+// @version      2022.3.31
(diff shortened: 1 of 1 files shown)


ROpdebee:
I've sent a message to the user that contacted me previously to check whether this fixes it. Regardless, thanks for the effort and giving this script some much-needed love!

ROpdebee:
Haven't heard back yet, but this LGTM.

PR #436, opened 2022-05-19
Minor updates for the cover art scripts

kellnerd:
Various minor improvements for ECAU and supercharged cover art edits, based on the [suggestions by chaban](https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/57?u=kellnerd) in the forums.

- Support new packaging types (one of them is non-square) and release statuses for supercharged cover art edits
- Use the latest version of Image Max URL for compatibility with new Violentmonkey (beta) versions (untested)
- Support filenames on archive.org which require URL encoding in ECAU:
  - New test for the given example to cover this case.
  - I'm restoring path separators after URL encoding to be on the safer side, although I haven't seen examples for full paths "including directory paths" yet, as per https://github.com/ROpdebee/mb-userscripts/blob/d04c2514a93e5a45a61cba0087a1161bf95c875d/src/mb_enhanced_cover_art_uploads/providers/archive.ts#L20

### DIFF
--- a/mb_supercharged_caa_edits.user.js
+++ b/mb_supercharged_caa_edits.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: Supercharged Cover Art Edits
-// @version      2021.10.21
+// @version      2022.5.19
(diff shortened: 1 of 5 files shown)


ROpdebee:
Thanks!

Issue #438, opened 2022-05-21
a-tisket seeder for Apple Music sometimes fails with status 429


kellnerd:
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/57

> On a rare occasion I saw a 429 from Apple. This caused the link for cover art seeding not to be added. Perhaps the script could gracefully retry or show a manual retry option?
> https://atisket.pulsewidth.org.uk/?spf_id=1WRqanANwPz8JyH0p78IPh
```
9999x9999-100.jpg:1          Failed to load resource: the server responded with a status of 429 ()
MB： Enhanced Cover Art Uploads.user.js:105074 [mb_enhanced_cover_art_uploads] Failed to add seed links to some cover art
onError @ MB： Enhanced Cover Art Uploads.user.js:105074
```

https://github.com/ROpdebee/mb-userscripts/blob/fa62e412476222c5dad9c73bf22b2ff856c99d78/src/mb_enhanced_cover_art_uploads/seeding/atisket.tsx#L76

Has to be related to the fetching of the image's dimensions since the insertion of the seed link itself should not cause an additional request to Apple's servers. The HTTP error makes the whole function fail, maybe these two features should be separated to avoid this.
Probably not worth to fix the underlying issue right now, automatic retries could be part of the refactoring of [mb_caa_dimensions](https://github.com/ROpdebee/mb-userscripts/blob/main/mb_caa_dimensions.user.js) into TS modules instead.

ROpdebee:
I think there's some function in the TS modules that allows for automated retrying with backoff, so fixing the underlying issue should indeed be part of a rewrite.

Would it make sense to move the image dimensions functionality on atisket to that script in its entirety? Ideally those results would be cached too, so I'd need to check whether we can share an IndexedDB cache between scripts.

PR #451, opened 2022-06-02
refactor(caa dims): migrate "MB: Display CAA image dimensions" to TypeScript

### LINKED ISSUE #439: Display image dimensions and size for RG cover art
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/57

> I don’t remember if it ever worked in either case but on “set cover art” pages neither dimension nor size are displayed and on release group pages size is always “??? KB”:
> - https://musicbrainz.org/release-group/e722cfc7-0975-4a33-82dd-bf37cc714321
> - https://musicbrainz.org/release-group/ab26a2e1-76bf-4007-ac86-8f064d456a8a/set-cover-art

ROpdebee:
It's finally time to migrate another script to TypeScript. I chose CAA dims because of some recent issues that need to be addressed there.

#### Current progress:
- [x] Rewrote the cache DB to use async/await.
- [x] Split cache into two stores, one for dimensions and one for file size. This allows us to store one if the other failed, which previously wasn't possible.
- [x] Added an `Image` implementation which glues together the dimension and file info loading and caching. This should also allow us to extend the implementation for a-tisket images in ECAU more nicely (we could possibly try and retrieve a file size for those images via a HEAD request). It also fixes part of #439, more specifically, about the unknown file size.
- [x] Refactor the image handling. I'm thinking a couple of classes that use composition to store an `Image` instance and also take care of displaying the information and deciding which `Image` subclass to use (queued upload thumbnails are handled differently to CAA images, so subclasses will enable us to easily distinguish between behaviours).
- [x] Clean up the CSS injection.
- [x] Decide what to do with the "exports" on the `window` object. We need one of those functions for Supercharged CAA Edits, and I'd like to keep backward compatibility (to avoid desync issues in upgrades).
- [x] Figure out the rest of #439, I'm guessing we're missing a query for the image on RG "set cover art" pages.
- [x] Add some retrying with backoff so we can later address #438 and integrate the new implementation into ECAU (separate PR, most likely).
- [x] Fix parsing of coverartarchive.org URLs in `Image.ts`, as Supercharged CAA Edits provides those URLs instead of direct IA URLs. If possible, we should also fix Supercharged to send direct IA URLs instead (or rewrite them in `Image.ts` to prevent the redirect).

---

### Bugs

- [x] Regression: https://musicbrainz.org/release/82af81c4-0cfa-4382-abc6-2ec08a79b431/reorder-cover-art (assertion error due to PDFs) (86781ca)

---

- [x] Via chaban:
> [File info is now centered](https://github.com/ROpdebee/mb-userscripts/commit/9bb74c60cf12fe454529fdb9bc8b162bd524d2e1#diff-0a9de830285fe85a4b5a85ff2394298b5473ca8790bb6c228558a4748fd8cb04R49-R53) but this doesn’t fit well on e.g. remove cover art pages.

https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/68?u=ropdebee
I fixed this locally but need to test some more before I push it. (cc7a357)

---

- [x] Another one via chaban (same post):
> When I first installed it no file info was shown. So I went and deleted localStorage and IndexedDB entries related to the script, afterwards it worked. Of course when I tried to reproduce it there were no problems.
¯\\_(ツ)_/¯

~Needs looking into.~ See below, same issue.

---

- [x] Build is failing with the new update notifications, since `index.ts` already imports `onDocumentLoaded` and the update notifications inject that import again, leading to errors. Also fixed locally but also needs more testing. (884e119)

---

- [x] Via chaban again:
> Script doesn’t run everywhere anymore. This looks intentional but some pages are missing like
https://musicbrainz.org/user/<username\>/edits/open or collections

(3111404)

---

- [x] When the script upgrades, it could be possible that another tab is still running the JS version with v1 of the cache DB open. That will block any upgrades to the schema and the cache promise won't resolve until the other tab is closed. We should probably instead fail the promise and fall back onto no caching at all in case it's blocked, and assume that the other tab will eventually be closed so that the schema upgrade can be carried out. Might be the cause of chaban's bug. (cf21e9a)

---

Fixes #439.


ROpdebee:
/deploy-preview I wonder whether it'll already work.

src/mb_caa_dimensions/style.tsx
```diff
@@ -0,0 +1,8 @@
+import USERSCRIPT_NAME from 'consts:userscript-name';
```
kellnerd, line 1: That's the constant you renamed into USERSCRIPT_ID in #458, so this has to be updated.

src/mb_caa_dimensions/style.tsx
```diff
@@ -0,0 +1,8 @@
...
+import css from './style.scss';
+
+export function setupStyle(): void {
+    document.head.append(<style id={'ROpdebee_' + USERSCRIPT_NAME}>
+        {css}
+    </style>);
+}
```
kellnerd, line 8: These style setup functions will probably appear in all userscripts sooner or later and a similar function also exists for the update notifications. On the one hand this cries for a shared function in `@lib`, on the other hand they are fairly trivial.

src/lib/util/format.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Format a number to a human-readable size string.
+ *
+ * @param      {number}  size    The size.
+ * @return     {string}  Human-readable, formatted version.
+ */
+export function formatSize(size: number): string {
```
kellnerd, line 7:
```suggestion
 * Format a number to a human-readable file size string.
 *
 * @param      {number}  size    The file's size in bytes.
 * @return     {string}  Human-readable, formatted version.
 */
export function formatFileSize(size: number): string {
```
Maybe a bit more specific?

src/mb_caa_dimensions/DisplayedImage.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+export class DisplayedQueuedUploadImage extends BaseDisplayedImage {
+    private readonly image: QueuedUploadImage;
+
+    // No cache, unnecessary to cache.
+    constructor(imgElement: HTMLImageElement) {
+        super(imgElement);
+        this.image = new QueuedUploadImage(imgElement);
+
+    }
```
kellnerd, line 182:
```suggestion
    }
```
Spacing around the first and last members of a class is also a bit inconsistent right now (sometimes there is an empty line, sometimes not), but that doesn't itch me as much as this unexpected empty line 😝

src/mb_caa_dimensions/InfoCache.ts
```diff
@@ -0,0 +1,268 @@
...
+            added_datetime: number;
+        };
+    };
+}
+
+type Migrator = (db: IDBPDatabase<CacheDBSchema>, tx: IDBPTransaction<CacheDBSchema, Array<StoreNames<CacheDBSchema>>, 'versionchange'>) => Promise<void>;
+/**
+ * Map of database migrations, indexed by current version.
+ */
+const DbMigrations: Record<number, Migrator> = {
```
kellnerd, line 236: This looks like an overkill to me, but since you've already written the code, there is no reason not to keep it 🤯

kellnerd (commented):
I have already installed the new version as soon as it was available and I didn't notice any regression so far. On the contrary, now it is finally working in all places where I'd expect it to 🙌

Ooof, that's almost a rewrite from scratch... Sorry, I'm afraid that I won't have time to check all of these commits 😇
Nevertheless, I still have left a few minor comments.

src/mb_caa_dimensions/InfoCache.ts
ROpdebee: It was simpler in the beginning 😅 I don't plan on doing many DB migrations anymore though.

src/mb_caa_dimensions/style.tsx
ROpdebee: It was actually renamed in #458, which isn't merged yet. So I guess we should wait until that is merged, merge `main` into here again, then do the necessary adjustments.

src/mb_caa_dimensions/style.tsx
ROpdebee: Yeah, I guess a function in `@lib` would make sense, but let's wait for #458 again so we can use that `@lib` function there too.

ROpdebee:
- Regression: https://musicbrainz.org/release/82af81c4-0cfa-4382-abc6-2ec08a79b431/reorder-cover-art (assertion error due to PDFs)

- Via chaban:
> [File info is now centered](https://github.com/ROpdebee/mb-userscripts/commit/9bb74c60cf12fe454529fdb9bc8b162bd524d2e1#diff-0a9de830285fe85a4b5a85ff2394298b5473ca8790bb6c228558a4748fd8cb04R49-R53) but this doesn’t fit well on e.g. remove cover art pages.

https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/68?u=ropdebee
I fixed this locally but need to test some more before I push it.

- Build is failing with the new update notifications, since `index.ts` already imports `onDocumentLoaded` and the update notifications inject that import again, leading to errors. Also fixed locally but also needs more testing.

ROpdebee:
/deploy-preview to incorporate new changes for our beta testers (thanks!)

kellnerd (approved):

PR #457, opened 2022-06-07
ci: automatically generate changelogs on deploy

### LINKED ISSUE #181: Automatically generate a changelog
One nice thing about Conventional Commits is that it would make it possible to automatically generate a changelog. I think there'd be some merit to having a changelog next to each userscript in the `dist` branch, which gets amended based on the PR title each time a release is made. Otherwise you need to go through each commit in that branch, which is a bit cumbersome.

ROpdebee:
Fixes #181.

There are various tools out there to automatically generate changelogs, but they're quite complicated and do much more than we need. So I wrote a simple custom one which should do the trick nicely.


build/deploy.ts
```diff
@@ -43,6 +44,8 @@ async function commitIfUpdated(scriptName: string): Promise<DeployedScript | und
 }
 
 async function commitUpdate(scriptName: string, version: string): Promise<DeployedScript> {
+    // Update the changelog
+    await updateChangelog(scriptName, version, distRepo, prInfo);
```
kellnerd, line 48: Do I understand this correctly, that the current code would only update the changelog for new PRs?
If that is the case, should we manually run this for already merged PRs and/or for all already existing commits in the `dist` branch?

ROpdebee: Yes, it'll only generate one entry for the current deployment, assuming the previous changelog was complete. Otherwise, generating a full changelog would require traversing the entire git history of the dist branch. I actually wasn't going to generate a changelog retroactively but it's probably a good idea to do that. I'll write up a quick script to generate the changelog for ECAU in the `dist` branch. Any changes prior to the dist branch should probably be excluded though, especially the ones where there were no conventional commits.

ROpdebee: https://github.com/ROpdebee/mb-userscripts/blob/dist/mb_enhanced_cover_art_uploads.changelog.md

kellnerd (approved):

PR #460, opened 2022-06-08
ci: fix finding PR number in deployment

ROpdebee:
I broke CI in #457, copy-paste error 🙁 Oops.

Via PR because I've restricted direct pushes to main and it's faster to submit a PR and merge it than to disable that restriction, push, and re-enable it. (Although it might've been faster considering I had to write all of this.) (And now that I wrote that, it definitely would've been faster…)


ROpdebee:
/deploy-preview as a test to see whether the `html_url` property is really present (according to the docs it is, but I didn't check).

PR #464, opened 2022-06-09
style: more linting and refactoring

ROpdebee:
Okay this turned out to be much larger than I anticipated. Whoops.

Pretty much just installed some new plugins and fixed the issues I agreed with and disabled those that I disagree with or are otherwise useless.

My initial intention was to made some stylistic parts more consistent (like the empty line issues discussed in some other recent PR, forgot which), so I checked `prettier`. Not a big fan of its output and restrictive ruleset though, so I've decided against it for now. Still on the lookout for another formatter. **Edit**: Turns out ESLint could do this all along.


ROpdebee:
Manually did some E2E tests, mainly focusing on the parts with non-trivial changes, and everything seems to work properly, so the unit tests did their job well 🙂

kellnerd (approved):
The new rules look sensible to me and I did a quick test to ensure that they are also working in my setup (by manually reverting some of the changes).

ROpdebee:
Thanks for the review @kellnerd. I'm going to hold off on merging these until `caa dims` is deployed though, otherwise I'd have to fix all these new warnings in that PR.

ROpdebee:
Actually, scratch that. If we delay the merging of this PR, we'd have to relint every other future PR as well.

PR #477, opened 2022-06-13
refactor(lib): allow tree-shaking the prototype input setter

ROpdebee:
https://github.com/ROpdebee/mb-userscripts/pull/473#issuecomment-1153312621

Probably not worth a new version, but `skip cd` will cause useless updates to be made to 2 of the 3 TS scripts on the next PR anway, and it'll bother me otherwise.

\+ A small fix for `split links` and refactoring to use `logFailure`, not worth separate PRs and deployments IMO.

### DIFF
--- a/src/lib/util/dom.ts
+++ b/src/lib/util/dom.ts
@@ -2,7 +2,7 @@
  * DOM utilities.
  */
 
-import { assertDefined, assertNonNull } from './assert';
+import { assertNonNull } from './assert';
 
 /**
  * Element.querySelector shorthand, query result required to exist.
@@ -70,14 +70,12 @@ export function parseDOM(html: string, baseUrl: string): Document {
     return doc;
 }
 
-// eslint-disable-next-line @typescript-eslint/unbound-method
-const nativeInputValueSetter = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value')?.set;
-assertDefined(nativeInputValueSetter);
+const inputValueSetterDescriptor = /*#__PURE__*/ Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value');
 
 // https://stackoverflow.com/a/46012210
 // Via kellnerd, https://github.com/kellnerd/musicbrainz-bookmarklets/blob/730ed0f96a81ef9bb239ed564f247bd68f84bee3/utils/dom/react.js
 export function setInputValue(input: HTMLInputElement, value: string, dispatchEvent = true): void {
-    nativeInputValueSetter!.call(input, value);
+    inputValueSetterDescriptor!.set!.call(input, value);
     if (dispatchEvent) {
         input.dispatchEvent(new Event('input', { bubbles: true }));
     }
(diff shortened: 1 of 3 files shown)


src/lib/util/dom.ts
```diff
@@ -70,14 +70,12 @@ export function parseDOM(html: string, baseUrl: string): Document {
     return doc;
 }
 
-// eslint-disable-next-line @typescript-eslint/unbound-method
-const nativeInputValueSetter = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value')?.set;
-assertDefined(nativeInputValueSetter);
+const inputValueSetterDescriptor = /*#__PURE__*/ Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value');
```
kellnerd, line 73:
```suggestion
const inputValueDescriptor = /*#__PURE__*/ Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value');
```

src/lib/util/dom.ts
```diff
@@ -70,14 +70,12 @@ export function parseDOM(html: string, baseUrl: string): Document {
...
 
 // https://stackoverflow.com/a/46012210
 // Via kellnerd, https://github.com/kellnerd/musicbrainz-bookmarklets/blob/730ed0f96a81ef9bb239ed564f247bd68f84bee3/utils/dom/react.js
 export function setInputValue(input: HTMLInputElement, value: string, dispatchEvent = true): void {
-    nativeInputValueSetter!.call(input, value);
+    inputValueSetterDescriptor!.set!.call(input, value);
```
kellnerd, line 78:
```suggestion
    inputValueDescriptor!.set!.call(input, value);
```

kellnerd (approved):
Yeah, I forgot to mention that rollup does not tree-shake computated constants by default, sorry. I had this problem a lot with my userscripts and bookmarklets and initially mitigated bundling of unused code by separating all functions which use one of these constants into their own module.
Then I discovered an option to make rollup assume that all of my [modules have no side-effects](https://github.com/kellnerd/musicbrainz-bookmarklets/blob/730ed0f96a81ef9bb239ed564f247bd68f84bee3/tools/buildUserscripts.js#L37) (will always be the case), which I preferred over scattering my code with `#__PURE__` comments. I don't know if you could also use that as I'm not 100% sure whether you have any (userscript) modules with side-effects.

ROpdebee:
> I don't know if you could also use that as I'm not 100% sure whether you have any (userscript) modules with side-effects.

The update notifications module (which we inject while building) is one with side effects. I'm also not sure about the 3d party dependencies that get bundled into the scripts, they might have side effects. While we could mark individual modules as having side effects in the rollup config, I'm a bit afraid that this could cause issues in the future (e.g. we add a new dependency which has side effects, like `core-js` polyfills, and rollup just eliminates the import).

In an ideal world, there'd be some TypeScript syntax to mark stuff as pure (like C++ `const` methods), which the TS compiler would verify. That information could be used in a bundler for tree-shaking, but alas, as far as I know, such syntax does not exist.

Issue #484, opened 2022-06-13
Script executing too early since MBS React changes


ROpdebee:
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/77?u=ropdebee

> This one doesn’t reliably work for me. Even for the same [release](https://musicbrainz.org/release/484db0c5-121a-49f3-9fa8-5b9481c676fd).
> 
> There is nothing in the console and API request goes through just fine. It was the only script enabled. ~Although when I enable some more it sometimes works.~ Apparently [INLINE STUFF](https://github.com/jesus2099/konami-command/blob/master/mb_INLINE-STUFF.user.js) makes it work.
>
> E.g. with this config it doesn’t work for me. But as soon as I enable INLINE STUFF it magically works.
> ```
> Display CAA image dimensions
> REVIVE DELETED EDITORS
> MASS MERGE RECORDINGS
> MERGE HELPOR 2
> Compare AcoustIDs easier! (enableMiniIcons = false)
> MASS ISRC
> Change release quality
> Mark recordings as video
> Lean display
> Seed the batch recording comments script
> Inline all recording's tracks on releases
> Set recording comments for a release
> ```
> 
> Once I also enable Blind Votes and Supercharged Cover Art Edits it stops working again. Plus I finally see an error:
> ```
> instrument.js:102 Loading “Thrivemix Presents: Electro” shadow release…
> instrument.js:102 Loading AcoustIDs…
> helpers.js:88 Uncaught TypeError: recAnchors is not iterable
>     at HTMLButtonElement.loadAndInsert (MB： QoL： Inline all recording＇s tracks on releases.user.js:95:20)
>     at HTMLButtonElement.dispatch (MB： Supercharged Cover Art Edits.user.js:4435:9)
>     at HTMLButtonElement.elemData.handle (MB： Supercharged Cover Art Edits.user.js:4121:28)
>     at HTMLButtonElement.r (helpers.js:71:23)
> helpers.js:88 Uncaught TypeError: recAnchors is not iterable
>     at HTMLButtonElement.loadAndInsert (MB： QoL： Inline all recording＇s tracks on releases.user.js:95:20)
>     at HTMLButtonElement.dispatch (MB： Supercharged Cover Art Edits.user.js:4435:9)
>     at HTMLButtonElement.elemData.handle (MB： Supercharged Cover Art Edits.user.js:4121:28)
>     at HTMLButtonElement.r (helpers.js:71:23)
[...]

PR #490, opened 2022-06-14
fix(inline recs): use MBS custom events to prevent hydration errors

ROpdebee:
Followup for #484. The previous fix in #489 works, but isn't ideal. Instead, we'll use the new custom events dispatched by MBS when hydration on an element finishes, introduced in https://github.com/metabrainz/musicbrainz-server/pull/2566.

Blocked until https://github.com/metabrainz/musicbrainz-server/pull/2566 gets released into production, but can currently be tested on test.musicbrainz.org.

### DIFF
--- a/mb_qol_inline_recording_tracks.user.js
+++ b/mb_qol_inline_recording_tracks.user.js
@@ -108,7 +108,7 @@ function loadAndInsert() {
 // listening for the correct one.
 function onReactHydrated(element, callback) {
     var alreadyHydrated = Object.keys(element).some(function (propertyName) {
-        return propertyName.indexOf('_reactListening') === 0 && element[propertyName];
+        return propertyName.startsWith('_reactListening') && element[propertyName];
     });
 
     if (alreadyHydrated) {


ROpdebee:
Updated to also handle the case when the userscript loads late and hydration already occured, per @jesus2099's comment in https://github.com/metabrainz/musicbrainz-server/pull/2566#issuecomment-1155740404.

I've also added a temporary check for the server version to decide between `load` events on the window and the `mb-hydration` events on the element, since these events now get dispatched in beta.mb.o. I think we should release this and let people test it in beta to ensure I'm not overlooking anything.

mb_qol_inline_recording_tracks.user.js
```diff
@@ -101,9 +101,28 @@ function loadAndInsert() {
...
-// to postpone initialising until cover art is loaded.
-window.addEventListener('load', () => {
+// MBS will fire a custom `mb-hydration` event whenever a React component gets
+// hydrated. We need to wait for hydration to complete before modifying the
+// component, React gets mad otherwise.
+// Multiple `mb-hydration` events will fire on a release page, so make sure we're
+// listening for the correct one.
+function onReactHydrated(element, callback) {
+    var alreadyHydrated = Object.keys(element).some(function (propertyName) {
+        return propertyName.indexOf('_reactListening') === 0 && element[propertyName];
```
kellnerd, line 111:
```suggestion
        return propertyName.startsWith('_reactListening') && element[propertyName];
```

kellnerd (approved):
Successfully tested on beta and prod.
(The injection of the button did also work on test, but clicking it causes an error because the MB API somehow did not return any recordings there... and the API is awfully slow.)

mb_qol_inline_recording_tracks.user.js
ROpdebee: Yeah, copy-pasted this from https://github.com/metabrainz/musicbrainz-server/pull/2566#issuecomment-1155799353 where I intentionally didn't use more modern browser APIs. I'm not sure about browser support for `String#startsWith`, but it should already be polyfilled by MBS.

PR #496, opened 2022-06-15
feat(caa dims): PDF dimensions and page count

### LINKED ISSUE #466: PDF dimensions
I think this is possible. During derivation, IA creates JPEGs for PDFs. It stores those in a ZIP, but provides transparent access to that ZIP. E.g. https://archive.org/download/mbid-82af81c4-0cfa-4382-abc6-2ec08a79b431/mbid-82af81c4-0cfa-4382-abc6-2ec08a79b431-29531260255_jp2.zip/

We could possibly grab the JPEG through there and load the JPEG's dimensions as a standin for the actual PDF's dimensions. Some further investigation would be necessary to ensure the dimensions are the same though.

Page count could likely also be extracted in the same way, and would be nice to know.

ROpdebee:
IA derives a ZIP of JPEG images for each page of a PDF. They also allow one to transparently access the contents of that ZIP. We can use this to grab the dimensions of each page. I'm going to limit this to just the first page for the time being, but I'll open a new issue to also handle the remaining pages. The reason why it's not in this PR is because I'm just not sure how to best display this information.

We're also calculating the number of pages in the PDF using the same ZIP transparency.

Closes #466.


kellnerd (commented):
Does the resolution of these derived JPEGs also say something about the PDF's quality/resolution or does it only depend on the size of the page? I'm just wondering whether this tells us anything beyond the aspect ratio of the first page (as I've never used the IA's derived images for a PDF).

ROpdebee:
I've checked one PDF and the JPEGs were the same dimensions as the PDF pages, so the dimensions should be the same as those of the PDF (at least its first page). The file size of the JPEG is of course not indicative of the quality of the PDF, but we don't use that (file size displayed is still the one from the original PDF file).

One thing I'm not sure of is whether this works for Text PDFs though, because I'm not sure whether IA does the same derivation for those and I haven't found an example of one of those yet. It's quite difficult to find any PDFs, really.

ROpdebee:
I can confirm it also works on Text PDFs.

PR #498, opened 2022-06-16
test(ecau): test for missing providers in docs

ROpdebee:
While adding new providers yesterday, I kept on forgetting to add them to the `supported_providers.md` file. This adds a test to verify that the new provider has an entry in that file.


kellnerd (approved):
Well, I don't fully understand how you use a mock to get the actual values/providers which would be added to the dispatch map, but if this works (and I can imagine how it does) that's a good addition 🧐
(I probably would have imported the already filled map from the provider module, but I can see that this approach would require to export the map just for this test.)

ROpdebee:
> I don't fully understand how you use a mock to get the actual values/providers which would be added to the dispatch map

We're mocking out the dispatch map, `providers/index.ts` does a whole bunch of `set`s on the dispatch map. All of those `set`s are logged, and then later we go through those calls and extract all the providers that were passed as arguments. In other words, we're using the mock to log which providers are available.

Exporting the map would've definitely be easier, but I don't think there's a way to easily iterate through all the providers in there. So we'd need to implement a method specifically for the tests, but it would also be unnecessarily included in the built scripts.

PR #499, opened 2022-06-16
refactor: more TypeScript migration

ROpdebee:
I'm not refactoring these a lot (yet), just migrating to TypeScript so we can benefit from all the benefits. Also removing jQuery where possible (mainly because I've already removed it from the dependencies and don't want to re-add it). Migrating to JSX where it makes sense too. And fixing linter warnings, of course.

Coverage is probably going to take a nosedive but it was a lie anyway. I've tested the new versions manually to ensure they still work.

- [x] `mb_blind_votes.user.js`
- [ ] `mb_bulk_copy_work_codes.user.js`
- [x] `mb_qol_inline_recording_tracks.user.js`
- [x] `mb_qol_seed_recording_disambiguation.user.js`
- [x] `mb_qol_select_all_update_recordings.user.js`
- [ ] `mb_supercharged_caa_edits.user.js`
- [ ] `mb_validate_work_codes.user.js`
- [ ] Update old scripts so they auto-update to the rewritten version.
- [ ] Update links in README.

I'll probably merge `validate work codes` into `bulk copy work codes` and rename the latter too.


kellnerd:
Just wanted to remember you, that there already is https://github.com/ROpdebee/mb-userscripts/tree/wip-blind-edits-rewrite

ROpdebee:
Yeah, that's a pretty big rewrite though. I'd like to get back to that eventually, but here I'm just focusing on refactoring them enough to get them to pass typechecking, linting, building and eventually deploying.

Also, I don't like bulk copy work codes' implementation. Very messy. 😅

ROpdebee:
I've superficially tested all of these, and tested Supercharged in a bit more depth because the changes were a bit more complicated there. Seems to work fine on my end, but I suggest leaving this open for a while longer to make sure there are no regressions. I'll probably post in the forums too to ask people to try out the previews and report any regressions.

ROpdebee:
/deploy-preview more fixes

So I switched to `@include` rules over `@match` rules because the new scripts are still running on too many pages. The `@exclude` rules weren't working (probably the domain part was bad?) and I've decided that I've had enough of the stupid `@match` ruleset, which is just.…_stupid_.

\<rant>

I _despise_ the new `@include` rules, though. I'm no regex hater (I actually quite enjoy them, maybe I'm weird), but it's just not something you want to "show" your users. If anyone now wants to know what pages these scripts are going to run on, they'll be presented with a string of alien civilization-level garbage. They're also much more error-prone (I wish I had addressed #75 before tackling this).

At the same time, the `@match` scheme imposes stupid limitations. I'm all for the safer domain matching, absolutely, but its path matching is severely limited. Why is there no complement to `@match` like there is `@exclude` for `@include`? Why is there no way to specify "one of these two options here please" without writing a near-identical rule twice? Why the f– does a wildcard (`*`) match a path separator (`/`)?! And WHY does it then NOT match the query string (`?...`) in Tampermonkey?!? I just…can't.

I just want to write `*.musicbrainz.org/release/*` without matching `musicbrainz.org/release/…/edit`. Why do I need to write `/https?://(\w+\?)?musicbrainz\.org/release/.+?([?#]|$)/` for such a simple match? 😕  Part of me wants to just say `@match *.musicbrainz.org/*` and add checks to the code to see if we're on a good page or not.

_For the record, I know this won't ever improve because there are different implementations and even if they all coordinate and decide to implement that (which is unlikely considering all the userscripts written so far likely expect `*` to match multiple path parts), the implementations will probably differ anyway and there's likely going to be this one person who cannot upgrade to the newer version and is still stuck on GM3, meaning we still have to support the older patterns anyway._

\</rant>

src/lib/util/metadata.ts
```diff
@@ -28,17 +28,105 @@ export interface UserscriptDefaultMetadata {
...
 /** Any pages on which edits can occur */
-export const MB_EDIT_PAGE_PATHS = [
-    'edit/*',
-    // <entity>/<entity_id>/edits, user/<username>/edits/open, search/edits?condition.0=..., ...
-    // TODO: This also matches /search/edits, on which no edits are shown. Should somehow be excluded
-    '*/edits*',
-    'user/*/votes',
-    '*/open_edits',
+export const MB_EDIT_DISPLAY_PAGE_PATTERNS = [
+    mb`edit/\d+`,
```
kellnerd, line 123:
```suggestion
    mb`edit/\d+`,
    mb`edit/(open|subscribed(_editors)?)`,
```
More pages with edit listings which I am regularly using.

src/mb_work_code_toolbox/index.ts
```diff
@@ -0,0 +1,786 @@
...
+    const editWorkFormQuery = 'form.edit-work';
+
+    function handleChange(): void {
+        // Whenever a change occurs, try to add buttons to the attr table.
+        // FIXME: We should check the changes in more detail before querying.
+        const workForms: Array<[HTMLFormElement, boolean]> = [...document.querySelectorAll<HTMLFormElement>(editWorkFormQuery)].map((f) => [f, false]);
+        document.querySelectorAll('iframe').forEach((iframe) => {
+            if (!iframe.contentWindow) return;
+
+            onWindowLoaded(() => {
```
ROpdebee, line 554: How the hell did this ever work in the rewritten version? We're asynchronously adding stuff to an array but then consuming the array synchronously 🤦

ROpdebee:
Noting for later: Supercharged is missing some separator on release events, c227924f4119639e917a47ed46afc82ea99628e7 wasn't a good fix.
> 2005-05-23 (GB)2005-06-06 (DE)

ROpdebee:
/deploy-preview let's see if this still works? We need to regenerate CAA Dimensions because the preview version was probably updated to a more recent main release while Supercharged is still expecting the new interface (it's very messy currently).

PR #502, opened 2022-06-19
feat(guess lang): script to guess language and script from tracklist

ROpdebee:
New script.

Language is guessed through [LibreTranslate](https://libretranslate.com/)'s detection API by feeding it all track titles + release name. We're using two open instances, first trying one that does very little (no?) rate limiting but has limited language support, then falling back on an instance that allows 15 calls per minute but supports more languages.

Script is guessed by counting the occurrences of character classes in the tracklist and release name. We're using Unicode-aware regular expressions and Unicode property escapes, babel transpiles these. It prefers non-Latin over Latin if both are present (but the non-Latin has to be at least somewhat common on the tracklist). All the "Frequently used" scripts should be supported (except Katakana. Han support is limited, see below).

### Known limitations and issues
- Tracklists with mixed languages aren't supported nicely. It cannot detect `[multiple languages]` and likely it'll produce no guess at all. This is a limitation in the API: 1) It only ever seems to return a single language; 2) We can't feasibly detect each track individually, as that would require too many requests.
- Script detection cannot distinguish between simplified and traditional Han and always uses the generic option. While it might be possible to improve these guesses, the only approaches to this I've seen so far literally list all of the characters that can appear in simplified/traditional, and those lists are huge. A heuristic based on release events might work (e.g. Taiwan is likely traditional).
- Although we could detect Katakana separately, we also fill it as Japanese, since the style guidelines say that Katakana should only be used for translations and 1) I don't think we can reliably detect that and 2) I'm not familiar enough with Japanese releases to implement that myself.
- Language support is somewhat limited. I've seen language detectors that claim they support 180+ languages, whereas the one we use supports maybe 40 at most. However, the other APIs require API keys and, more importantly, only allow like 50 requests per month on a free plan.
- Both script and language detection can, of course, produce false matches. Such is life.
- I'm not a huge fan of the button position, I'd rather put a "Guess case"-like button next to one of the input fields, but there are two input fields and it'd be weird to only have a button next to one of them.
- Some configuration options might be desirable, e.g. setting the minimum confidence thresholds.

### DIFF
--- a/src/lib/MB/types.ts
+++ b/src/lib/MB/types.ts
@@ -18,10 +18,27 @@ export interface ExternalLinks {
     };
 }
 
+export interface ReleaseEditorMedium {
+    loaded(): boolean;
+    loading(): boolean;
+    loadTracks(): void;
+    tracks(): Array<{
+        name(): string;
+    }>;
+}
+
+export interface ReleaseEditorFields {
+    release(): {
+        name(): string;
+        mediums(): ReleaseEditorMedium[];
+    };
+}
+
 export interface ReleaseEditor {
     externalLinks: {
         current: ExternalLinks;
     };
+    rootField: ReleaseEditorFields;
 }
 
 declare global {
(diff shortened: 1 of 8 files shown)


ROpdebee:
chaban has a bunch of feedback at https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/90?u=ropdebee which I still need to process.

PR #504, opened 2022-06-21
fix(split links): run in iframes in Chrome

ROpdebee:
Fixes issue reported by [chaban in the forums](https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/84?u=ropdebee).

We already knew that Firefox doesn't fire the `DOMContentLoaded` event, which is why we were listening to the iframe window's `load` event. However, it turns out that Chrome doesn't fire that event. I also tried mutation observers, they don't work either. I'm not sure whether Chrome fires the `DOMContentLoaded` event, but considering that the iframe's document `readyState` attribute is set to `'complete'` immediately makes me doubt that. So we'll resort to the simplest approach: Continuous polling (eww!).

I'm not 100% certain the issue chaban is seeing, but I couldn't reproduce it on Firefox, and it was broken in Chrome either way, so it needed fixing. I'll ask chaban to test this as well to verify that the problem is solved.

We'll also need to apply this to `Work code toolbox` in #499.


ROpdebee:
/deploy-preview again, better approach. Not sure where my head is at these days.

ROpdebee:
I think chaban confirmed that the last attempt did fix the problem. This fix should go into the work code toolbox in #499 too.

PR #512, opened 2022-07-03
fix(split links): adjust links editor properties for new MBS version

ROpdebee:
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/91

https://github.com/metabrainz/musicbrainz-server/pull/2497 changed some internals that we were using to get the external links editor object (specifically, https://github.com/metabrainz/musicbrainz-server/pull/2497/files#diff-20a3affccc96c36d0a67c0ca98ebc7010590a1adc067eeea57147b0302bf1add).

This is only in beta for now, so for now we'll try both the fixed and old implementation and use the one that worked, but once this goes into prod we should remove the old implementation.

### DIFF
--- a/src/lib/MB/types.ts
+++ b/src/lib/MB/types.ts
@@ -20,7 +20,9 @@ export interface ExternalLinks {
 
 export interface ReleaseEditor {
     externalLinks: {
-        current: ExternalLinks;
+        externalLinksEditorRef: {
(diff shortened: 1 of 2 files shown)


ROpdebee:
Since this has now landed in prod and it's a minor change, I'll merge without review.

PR #516, opened 2022-07-03
feat(ecau): progress indicators

ROpdebee:
Fetching and queueing images can take a long time, depending on the number of images there are and their size. I've added a bunch of progress indicators as "waiting therapy". Summary:
* Providers with track images (mainly Bandcamp) can take a while to find and pre-process all track images, so I've changed the logging to also emit a message whenever an image is processed. It'll now say e.g. “Checking for track images, this can take a while… (3/9)”, incrementing as each track image gets loaded.
* Instead of waiting until the end to enqueue all images at once, we'll now enqueue them as soon as they're ready.
* While fetching an image, a progressbar is now shown tracking the download progress.
* We'll also now block submissions of the edits until all input URLs are fully processed. Otherwise, it'd be possible for users to submit an edit while ECAU is still actively loading and enqueueing images.

Also includes some behind-the-scenes changes to make this possible.


ROpdebee:
/deploy-preview for the lurkers who want to get it early. I know you're watching :eyes:

ROpdebee:
YOLO

Issue #526, opened 2022-07-18
Not all Soundcloud track images are fetched


ROpdebee:
Seems like Soundcloud added pagination and track images are limited to the first 5 tracks.
Example: https://soundcloud.com/schnozz/sets/berdlycore-chapter-2

PR #543, opened 2022-08-03
feat(ecau): extract Soundcloud backdrop images

ROpdebee:
While fixing #526 I noticed that the Soundcloud tracks in that set have backdrop images, so let's extract those too.
Also included is a fix for #526. Not sure if this problem has always existed or was caused by a Soundcloud change, because we didn't have tests for sets with >5 tracks.


src/mb_enhanced_cover_art_uploads/providers/soundcloud.ts
```diff
@@ -120,16 +150,77 @@ export class SoundcloudProvider extends ProviderWithTrackImages {
         // Don't bother extracting track covers if they won't be used anyway
         if (onlyFront) return covers;
 
-        const trackCovers = filterNonNull(metadata.data.tracks
+        // Soundcloud only loads data for the first 5 tracks at first, need to
+        // load the rest of the tracks later.
```
kellnerd, line 154: I assume later = now, i.e. in the next line?

src/mb_enhanced_cover_art_uploads/providers/base.ts
```diff
@@ -116,6 +116,7 @@ export abstract class CoverArtProvider {
 export interface ParsedTrackImage {
     url: string;
     trackNumber?: string;
+    customComment?: [string, string];  // singular, plural
```
kellnerd, line 119:
```suggestion
    customCommentPrefix?: [string, string];  // singular, plural
```
Makes it clearer whether we expect a comment prefix or a full comment which will be used verbatim.

ROpdebee: Good point. I'm sure you'll find loads of other poorly-named variables/properties in the changes over the last couple of months 😅

kellnerd (approved):
Looking good, but I did not test 😇

PR #544, opened 2022-08-03
test(ecau): fix tests after upstream provider changes

### LINKED ISSUE #509: Re-record PollyJS HTTP recordings
the tests.

# Why?

# How?

Thank you!

ROpdebee:
No changes to prod code necessary, but some upstream changes are causing test failures when recordings are regenerated. Some of them were already known previously (#430).

* The Deezer release we were using to test placeholder filtering no longer has a placeholder. Changing to a replacement.
* Spotify now returns a 404 status code instead of 200 on releases that don't exist. The custom check wouldn't be needed anymore, but let's keep it in case they change it to 200 again.
* Jamendo cover art ID changed (but it's still the same image, not sure what happened there).
* 2 Audible releases got updated with new cover art, tests have been adapted to check against those new URLs.

I also re-generated the remainder of the recordings, but they didn't cause any test failures, so as per usual I'm not going to update them in the repo.

Closes #509.


kellnerd:
Well, I should probably have chosen less popular Audible releases because I knew that these have already been reissued with different cover art a few times in the past 😂

PR #546, opened 2022-08-04
feat(ecau): rich copy-paste of webpage images and links

ROpdebee:
We now also accept input from copied web pages and parse URLs from anchors and images on the copied portion, which should make it easier to copy images from providers which are currently not supported (and might never be, depending on their popularity).

Some demos because a picture says a thousand words:

https://user-images.githubusercontent.com/15186467/182946797-b658a3e8-8536-41ab-aad8-0c861422308f.mp4

https://user-images.githubusercontent.com/15186467/182947022-71e21fc1-c562-47ae-b94a-b004462ac131.mp4

The ebay demo is a bit weird because it doesn't highlight the selection, but the copied portion is just the two sidebar images and not the full-page selection that appeared earlier in the clip.

This probably needs a forum post after release because the feature notification won't properly explain what this is.


ROpdebee:
/deploy-preview for beta testing because the extraction may not be ideal yet and the easiest way to find areas for improvement is by having people use the new version 🙂

kellnerd (commented):
I've tested this feature a bit and it's a great thing to have (as a substitute for a generic provider). There are two things which don't bother me personally but might confuse the casual user:
- The pasted HTML fragment may contain irrelevant weblinks which are then of course leading to error messages (unsupported provider etc.). Personally I would have mitigated that by only extracting URLs from `` tags, but your demo video shows a use case where the extraction from `<a>` tags makes perfect sense. Maybe anchors could be an optional feature or we could ignore them unless there are no `` tags present in the HTML fragment?
- Every image is handled as separate input, so the final status message (in case of success) will always be "Successfully added 1 image(s)", no matter how many images were actually extracted.

ROpdebee:
/deploy-preview with new fixes

> Maybe anchors could be an optional feature or we could ignore them unless there are no  tags present in the HTML fragment?

I've also been thinking of such a system. `` first, then `<a>`, then matching URLs in plain `textContent` (in case someone copies unformatted links from a webpage. Should be fixed now.

> Every image is handled as separate input, so the final status message (in case of success) will always be "Successfully added 1 image(s)", no matter how many images were actually extracted.

That was actually already the case for multiple URLs split by whitespace. I thought I changed that when adding more progress indicators, but apparently I forgot. Should also be fixed now.

ROpdebee:
/deploy-preview it's been over half a year, I think it's about time to release this

Issue #521, opened 2022-07-14
Failed to grab images: Failed to extract images from embedded JS on generic physical page


HibiscusKazeneko:
I've been getting this error message fairly frequently when attempting to import cover art from Amazon pages. It seems to be an issue with newer listings.

ROpdebee:
Do you have an example of a URL on which it fails? That'll make it much easier to figure out what the issue is. Thanks!

HibiscusKazeneko:
https://www.amazon.com/gp/product/B0B2F6TV5P was affected at the time I posted this thread. It appears to have fixed itself.

ROpdebee:
Hmm, strange that it fixed itself. It could be that the script landed on a captcha page due to rate limiting etc. Someone (probably me) should investigate to see 1) if we can figure out why it's actually failing to parse the page and 2) if we can either fix it or improve the error message if it turns out to be unfixable.

HibiscusKazeneko:
This issue is back with a vengeance. For the past several hours I have received this exact error message every time I try to import cover art for a physical release. It's the same for older and newer releases.

ROpdebee:
I've pushed an update that'll improve the error message in case it hits rate limiting with a captcha. I'm not sure that it's caused by captchas though, because I haven't been able to reproduce it yet.

I've also got some work in progress that should improve the "debugability" of the script in case it hits weird errors. It needs a bit more work to be usable, and I can't really give an ETA on when it'll land, but when it does, I'll be able to look at this in more detail.

PR #549, opened 2022-08-08
refactor(ecau): standardise 404 error messages for providers

ROpdebee:
Previously, when a release returned a 404 code, a generic HTTP error would be shown in the log. Changing this to be more specific and to include the provider name in the error message.

Not all providers have been changed though. Notably the Tidal provider will still throw a generic error because it performs custom requests. That remains to be fixed.

### DIFF
--- a/tests/unit/mb_enhanced_cover_art_uploads/providers/tidal.test.ts
+++ b/tests/unit/mb_enhanced_cover_art_uploads/providers/tidal.test.ts
@@ -85,8 +85,6 @@ describe('tidal provider', () => {
         const extractionFailedCases = [{
             desc: 'non-existent release',
             url: 'https://listen.tidal.com/album/1',
-            // FIXME: Tidal provider doesn't use the base `fetchPage` and only throws a generic error.
-            errorMessage: 'HTTP error 404: Not Found',
         }];
 
         // eslint-disable-next-line jest/require-hook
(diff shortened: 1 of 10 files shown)


tests/unit/mb_enhanced_cover_art_uploads/providers/tidal.test.ts
```diff
@@ -85,6 +85,8 @@ describe('tidal provider', () => {
         const extractionFailedCases = [{
             desc: 'non-existent release',
             url: 'https://listen.tidal.com/album/1',
+            // FIXME: Tidal provider doesn't use the base `fetchPage` and only throws a generic error.
```
kellnerd, line 88: Is this a non-trivial fix or do you simply haven't looked at it yet?

kellnerd (approved):

tests/unit/mb_enhanced_cover_art_uploads/providers/tidal.test.ts
ROpdebee: It's easy enough to fix by wrapping the custom request in a try-catch, but that'd be another case of
```
let response;
try {
    response = await ...
} catch ...
```
and that pattern is starting to annoy me…

What I'd like to do instead is add some sort of optional "error factory" to `gmxhr` to create custom errors based on some conditions but that's a less-trivial change. Moreover, I've got some other ideas in the pipeline (logging responses somewhere so that they can be included in a bug report to debug e.g. #521), and for that I was thinking of having a unified interface for `gmxhr` and `fetch` so that both can be logged (not essential, but it'd be nice to be able to use them in the same way). So IMO it'd make sense to group those changes together. TBH it'd make sense to group those changes into this PR as well.

ROpdebee: The error message issue has been addressed, the interface unification is something I'm still considering and will go into the PR with improved logging if I decide that the unification is necessary.

kellnerd (approved):

Issue #569, opened 2022-09-23
Supercharged Cover Art Edits is breaking inline entity submission


kellnerd:
Today I was no longer able to create a new artist inline in the release relationship editor (main server). The submit button of the artist modal was always disabled until I disabled this specific userscript (by trial and error).
Although this script should never run on this page to begin with (as we both already know), this has not been a problem so far. I guess VM 2.1.3.1 introduced this issue as the script started to appear in a new section "Sub-frames only scripts":
![screenshot](https://user-images.githubusercontent.com/52860029/191982116-de6cdbdd-cb81-40d8-b8f2-0303a13cc3a8.png)
To mitigate the problem, I've currently added a custom exclude-match rule in the VM settings: `*://*.musicbrainz.org/dialog*`

PR #574, opened 2022-10-02
fix(caa-edits): Exclude inline entity creation dialogs

kellnerd:
The script was also running in sub-frames since Violentmonkey 2.13.1, which prevented inline entity dialogs from being submitted.
Temporary fix for #569 until we have better match rules for this script.


ROpdebee:
Thanks!

HibiscusKazeneko:
Unfortunately it doesn't appear to have solved the problem :/

kellnerd:
@HibiscusKazeneko Version 2022.10.2 of this userscript has solved the problem for the selection of userscripts which I have installed. Maybe you have other userscripts installed which are also affected by the change of Violentmonkey's behavior? I think you can find possible candidates in the "Sub-frames only scripts" section of the VM menu, see #569.

Another thing which you could try (if you still think this userscript is the cause) is to use `@noframes` instead of `@exclude-match *://*.musicbrainz.org/dialog*`. This approach also works for me, but I'm not 100% sure whether there aren't any frames where the userscript is supposed to run.

Issue #547, opened 2022-08-06
[MB: Enhanced Cover Art Uploads] Add support for Bugs!


Maxr1998:
[Bugs!](music.bugs.co.kr) is a Korean music streaming and purchase site. Provides very high-quality cover arts, good alternative to Melon.

[Example page](https://music.bugs.co.kr/album/4078166). The `og:image` tag contains a link to the cover art (https://image.bugsm.co.kr/album/images/500/40781/4078166.jpg), for the highest quality the `500` can be replaced with `original`: https://image.bugsm.co.kr/album/images/original/40781/4078166.jpg.

Maxr1998:
Thanks for the quick implementation!

Unfortunately, it seems like it broke after a short while. Could you look into it?

ROpdebee:
The example release you linked appears to work for me, as well as some other albums I found on the main page.

Do you have an example of a release that no longer works?

Maxr1998:
I realized it's not really an issue with the userscript, but device related. I'm using the script with Kiwi Browser on Android, and it uses the mobile m.bugs.co.kr site there when fetching the album art.

Unfortunately, the `og:image` tag is slightly broken on the mobile site, since it omits the protocol part of the URL:

```html
<meta property="og:image" content="//image.bugsm.co.kr/album/images/120/40797/4079793.jpg"/>
```

This causes the following error in script:

```
mb_enhanced_cover_art_uploads] Failed to fetch or enqueue images TypeError: Failed to construct 'URL': Invalid URL
    at userscript.html?name…f0c3a3e30:106255:18
```

To work around this issue, you could either hardcode a useragent when fetching the site or handle a missing protocol by prepending it manually.

ROpdebee:
Thanks for debugging this. I've pushed a potential fix in #581 but I'm unable to test this on a mobile device without lots of hassle, so it'd help a lot if you could try out the preview version in https://github.com/ROpdebee/mb-userscripts/pull/581#issuecomment-1272587335 to see if the fix works.

Maxr1998:
Unfortunately not. The image URL in the meta tag uses image.bugsm.co.kr as the domain, whereas the site URL is either music.bugs.co.kr or m.bugs.co.kr. The image URL on the mobile site isn't really relative, just broken due to the missing protocol, unfortunately.

ROpdebee:
[...]

PR #591, opened 2022-10-15
feat(ecau): add ability to export debug logs and HTTP responses for bug reports

ROpdebee:
It's sometimes quite difficult to reproduce an ECAU issue, especially when it concerns specific providers. #521 and https://github.com/ROpdebee/mb-userscripts/issues/547#issuecomment-1272552566 come to mind. Moreover, we're doing some debug logging here and there, but those logs are not shown to the user, so we won't get them in bug reports.

These changes add a feature that enables users to download a log file for an ECAU run, including debug messages and provider responses. These could then be submitted alongside a bug report in case we can't reproduce an issue. When exporting the log, it also displays a warning about the domains in the recorded responses so that people don't accidentally leak user credentials that may be stored in the response.

We're also switching to a new request interface that unifies `fetch` and `GM.xmlHttpRequest`, which was necessary to log the responses.


ROpdebee:
/deploy-preview just to make sure deployments will still work

PR #601, opened 2022-10-23
ci(deploy): check for changes against base instead of previous deploy

### LINKED ISSUE #600: `skip cd` tag inefficiencies
The `skip cd` tag on PRs can be used to prevent deploying updates when it's undesired to push an update. For example, #593 and #598 cause changes to userscripts which are not worth deploying an update for.

However, merging these with `skip cd` wouldn't prevent the meaningful updates entirely, it would only prevent them from being pushed immediately when the PR is merged. If another change happens later on that doesn't involve changing the scripts (e.g. updating a dev dependency), the changes would still be deployed. In those cases, the changelogs would also be wrong. Instead, the updates should be postponed until the next change that actually changes the userscripts.

In other words: Deployment automation currently fires when 1) there is no `skip cd` tag and 2) the built scripts are different from the deployed scripts.
Instead, it should be changed to 1) there is no `skip cd` tag and 2) the built scripts are different from the last version before the change (even if that last version wasn't deployed).

ROpdebee:
The continuous deployment checks whether changes have been made to the scripts and deploys an update if so. However, it was checking against the latest deployed version, rather than the "previous version". These two are different if the `skip cd` tag is used. This led the deployment to deploy updates for PRs that did not change the scripts at all, because an old, not-yet-deployed change was skipped previously.

Instead, we'll now check against the compiled scripts based on the "previous" commit ref. For pushes to main, this is the commit before the push (even when multiple commits are pushed); for deployment previews in PRs, this is the base branch of the PR; for checking for changes in automated dependency updates, this is the main branch.

This should fix #600.

To test this, we'll merge #593 and #598 with `skip cd` tags, generate a deployment preview for #580 (which should cause no deployments), and merge #580, which also shouldn't cause a deployment.

### DIFF
--- a/build/check-scripts-unchanged.ts
+++ b/build/check-scripts-unchanged.ts
@@ -20,7 +20,7 @@ async function checkUserscriptsChanged(): Promise<void> {
             throw new Error('I encountered a userscript which has not been deployed yet!');
         }
 
-        // Check against the main branch's HEAD.
+        // Check against the main branch.
         if (await userscriptHasChanged(scriptName, 'main')) {
             throw new Error(`Userscript ${scriptName} would be changed`);
         }
--- a/build/check-scripts-unchanged.ts
+++ b/build/check-scripts-unchanged.ts
@@ -20,7 +20,8 @@ async function checkUserscriptsChanged(): Promise<void> {
             throw new Error('I encountered a userscript which has not been deployed yet!');
         }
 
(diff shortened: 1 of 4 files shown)


ROpdebee:
Terribly sorry to anyone who is watching this repository and getting all of the spammy emails because of CI changes. 😢

PR #602, opened 2022-10-23
ci(deploy): fetch full repository history

ROpdebee:
We were only fetching at depth 1, which meant we couldn't check out an older commit to compare compiled scripts during deployment.

_Seems like every time I change something about CI, I need to do a follow-up PR to fix a bug in CI 🤔_


ROpdebee:
_UGHHHHH_

PR #612, opened 2022-11-06
fix(ecau): gracefully reject Amazon Music pages

### LINKED ISSUE #611: Amazon Digital Music pages fail to extract
So Amazon overhauled their digital music pages on the main site. E.g. https://www.amazon.com/dp/B07R92TVWN. Everything is dynamically loaded through an API now (https://na.mesk.skill.music.a2z.com/api/showHome), with (seemingly) no way of grabbing cover art without also going through the API. The API appears to block any request that doesn't include like 20 custom headers, also requiring a bunch of unique data  (CSRF tokens, request IDs, session IDs, etc). At first glance it looks possible to create these headers, since it appears that unique data is available in the HTML page that we already load. However, the images that the API returns are limited to 500x500, and higher resolution images like the ones we were getting previously don't seem to be served anymore. Thus, I'm on the fence about whether the required effort to support these new pages would even be worth it, assuming that most (if not all) Amazon Music releases are available on other streaming services with higher resolution cover art.

Some alternative solutions:
* Use a seeding button like on a-tisket for those pages.
* Do some trickery with `window.open` and `window.postMessage` so that ECAU lets a new tab/popup load the image, runs on that tab, and communicates the image URL back to the MB page.

TL;DR: Amazon has made it very difficult to extract cover art from Amazon Music releases, and the artwork is of such low quality that I doubt that adding support for those pages is worth it. Maybe we should just fail gracefully (instead of "Failed to extract images from generic page") and tell the user to use a different source.

Since I don't use Amazon as a source a lot, I'm not sure whether it's generally true that these releases are available elsewhere or whether it's even acceptable to drop support for Amazon Music. @hibiscuskazeneko and @kellnerd, do you have any insights? Or do you possibly know of other means of getting these images?

ROpdebee:
Amazon made it very difficult to extract cover art from these pages, so for now, we'll drop support for them. See #611.

Closes #611.


kellnerd (approved):
LGTM, a quick test with the example release https://www.amazon.com/dp/B07R92TVWN works as expected.

In an additional test I noticed that our translation of music.amazon.* URLs is not foolproof, as the releases might have different ASINs on Amazon Music, e.g. https://music.amazon.com/albums/B07RC3958J for the above example. While the first ASIN automatically redirects to the second ASIN on Amazon Music, this does not work the other way around on the main domain, resulting in 404 errors ("Amazon release does not exist", which is a bit misleading in this case).
Not sure if this has always been the case, but if you don't have an example of an Amazon Music release with a working redirect, we should directly throw an error in the Amazon Music provider instead of passing the request to the Amazon provider first.

ROpdebee:
It doesn't really make sense to do the translation anymore, since it'll translate to the unsupported page anyway, so let's just always throw there too. I've opted to throw the error rather than outright removing the provider, as I don't want to rule out the possibility of re-adding support if the situation changes in the future.

kellnerd:
Yeah, that makes sense. The only thing which is left now is updating `supported_providers.md`.

PR #661, opened 2023-04-22
feat(ecau): add NetEase/163.com provider

### LINKED ISSUE #645: ECAU: Support NetEase (music.163.com) urls for import
Sometimes NetEase is the only source url for releases, so it would be nice to support importing cover art from them.
Given a release url at:
`https://music.163.com/#/album?id={id}`

Album art can be found via the api:
`https://music.163.com/api/album/{id}`
In json object as `obj.album.picUrl`

Or via the page source:
`document.querySelector(".j-img").getAttribute('data-src');`

ROpdebee:
Fixes #645.


ROpdebee:
/deploy-preview because I need to test this in MB but I'm too lazy to build locally and copy-paste the new script.

PR #662, opened 2023-04-22
feat(ecau): add Juno Download provider

### LINKED ISSUE #658: Juno download
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/124?u=ropdebee

ROpdebee:
Easy one. Fixes #658.


ROpdebee:
There's also the sister site juno.co.uk (https://en.wikipedia.org/wiki/Juno_Records) but I think that should be turned into a separate provider, since it also offers physical releases and the HTML is likely slightly different. Even if the HTML was the same, I'd still prefer putting them into two separate buttons with different names and different icons to avoid confusion.

PR #663, opened 2023-04-23
feat(ecau): add Booth.pm provider

### LINKED ISSUE #660: Booth.pm
https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/126?u=ropdebee

> Another possible website for ECAU: https://booth.pm/en/browse/Music . If it could extract covers from here like it can from itunes that would be great for physical CDs from japan. Example: https://musicbrainz.org/release/10c4f2ca-a230-4e73-9985-7ec0ffc90f82
>
> I have no idea if this would be possible but worth asking I suppose.

https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/127?u=ropdebee
> But that booth.pm example mixes 3 different releases on the same page with only one cover (supposedly the download edition cover art, only):
>
>  *  Booth “another jacket” edition (I think it means alternate front cover)
>  *  Download edition
>  *  CD edition with additional sleeve (probably non-square front cover)

https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/128?u=ropdebee
>I think it’s two editions, limited edition with “another jacket” and another with a bonus exclusive mp3 download you get after ordering. I added the link to that release before I noticed that bonus, made a new release for it and removed that link but it’s still pending. So only one release for sale there currently.
>
>A better example: https://musicbrainz.org/release/90069298-34f4-4d79-9b67-6d142e071549

https://booth.pm/ja/items/1973472
https://booth.pm/en/items/2969400

ROpdebee:
Fixes #660.


ROpdebee:
We should probably squash this one when merging, primarily because of the page HTML responses included in earlier commits that are now replaced by a (much smaller) JSON.

PR #686, opened 2023-07-02
feat(split links): add bulk URL copy buttons to a-tisket

ROpdebee:
The originally created the script to be able to quickly add the links provided by a-tisket to an existing release. These buttons make it more convenient to copy them.


ROpdebee:
/deploy-preview

Currently a draft because I feel like this needs a bit of refactoring so that we put the a-tisket and the MB stuff into different files. That way we could likely also get rid of the ugly IIFE that was added. I'll get to that later™. Nonetheless, it seems to work, so lurkers, feel free to use the pre-release version for now.

PR #687, opened 2023-07-12
fix(ecau): prevent image blob from unloading before upload

ROpdebee:
It appears that Violentmonkey's implementation of `GM.xhr` with a blob `responseType` causes the blob to become unloaded if the response itself goes out of scope. When that happens, the image we enqueued for uploading will become empty, and the upload will fail. To prevent this from happening, we'll instead copy the buffer, which should prevent the contents from being unloaded.

### DIFF
--- a/src/lib/util/blob.ts
+++ b/src/lib/util/blob.ts
@@ -8,29 +8,22 @@ function hexEncode(buffer: ArrayBuffer): string {
         .join('');
 }
 
-export function blobToDigest(blob: Blob): Promise<string> {
-    async function onLoad(reader: FileReader): Promise<string> {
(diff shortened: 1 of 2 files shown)


ROpdebee:
/deploy-preview for more extensive testing

I've done a few quick tests replicating circumstances in which the issue would appear 90% of the time for me, and this does appear to fix it. Still, I'd like to test a bit more before releasing.

ROpdebee:
I've tested this for several days under everyday editing circumstances and while the issue would occur quite regularly before, it hasn't occurred at all with these changes, so I'm quite confident this fixes it.

PR #689, opened 2023-07-19
fix(ecau): upstream provider fixes

### LINKED ISSUE #665: Re-record PollyJS HTTP recordings
the tests.

# Why?

# How?

Thank you!

### LINKED ISSUE #688: Failed to grab images from Beatport
The Beatport made its website change recently, and the cover art uploader does not work.

ROpdebee:
- Fix issue with new recordings that duplicate responses.
- Fix Beatport provider following site changes (fixes #688)
- Fix Amazon test case and clean up Amazon Music recording which is now useless.
- Fix Soundcloud Client ID extraction, it's changing too frequently so we need to automate the extraction. Should also be somewhat robust if the client ID changes again in the future by automatically trying to refresh it. Not everything is covered by tests, but let's ignore that for now.

Closes #665.


ROpdebee:
Forgot to mention:
- DatPiff seems to be down and instead advises people to go to archive.org. There's no automated redirects in place, and it's a bit unclear what'll happen to the old releases, so let's wait for more clarity until we decide what to do with the provider.
- The Booth.pm item in the test case for missing artwork is 404, but I couldn't be bothered to look for a new one yet.

PR #710, opened 2023-11-30
fix(ecau): update Discogs GraphQL query hash

### LINKED ISSUE #709: [Enhanced Cover Art Uploads] Broken import from Discogs?
Recently I'm getting HTTP error 403 trying to fetch covers on multiple albums from Discogs. May be temporary issue but reporting this to have track about it.

kellnerd:
The old hash was leading to 403 errors, so I've manually extracted the new one.
Should fix #709 for now.


ROpdebee:
Seems like the query hash is embedded in `https://catalog-assets.discogs.com/main.[HASH].js`. Strange how it already changed though, perhaps it's dynamically regenerated?

kellnerd:
No, I've just been stupid and copied the wrong hash which was for a different API endpoint.
Now I've finally updated my local dev setup and successfully tested with a different hash.

/deploy-preview again please

kellnerd:
Thanks, you were faster than me with updating the test data 😀
I still had to update Node.js and the npm dependencies to make the tests work again locally (after I last ran them about a year ago).

ROpdebee:
For future reference: The Discogs official API returns images but only when the user is authenticated. The images also seem to be slightly higher resolution than the ones we get from the internal API (e.g., 600x649 instead of 554x599), but they're still not full-size. But worth considering in the future perhaps, if we can get users to be authenticated.

Thanks @kellnerd!

PR #711, opened 2023-11-30
fix: upstream provider fixes

### LINKED ISSUE #697: Re-record PollyJS HTTP recordings
the tests.

# Why?

# How?

Thank you!

ROpdebee:
* Audiomack new layout
* Updated test URLs for booth.pm

Unfixed:
* DatPiff site is still down
* Amazon Audible audiobook redirection is broken. If I recall correctly, if we were given a product page for an Audible audiobook, we'd rewrite to an Audible-specific page to get higher resolution images. However, that URL rewrite now leads back to the Amazon main page, so we can't use that special case anymore. @kellnerd do you know what the new Audible URL should be (or of any other tricks)?

Closes #697


kellnerd:
> However, that URL rewrite now leads back to the Amazon main page, so we can't use that special case anymore.

In which scenario does it redirect for you?
For me these `https://www.audible.com/hz/audible/mlp/mfpdp/` URLs are simply "Page not found" error pages now and in addition the ASINs for Audible are no longer the same as on Amazon. Our test case https://www.amazon.com/dp/B017WJ5PR4 corresponds to https://www.audible.com/pd/B017WJ5ZK6 for example.
Since that page also offers only the lousy 500x500 cover image we could skip this special handling completely. Unfortunately I haven't been able to extract any hires covers from Audible (or Amazon digital audiobooks) as of today (and haven't used it as a source for many months), so it might be no longer possible...

ROpdebee:
I think we used to rewrite `https://www.amazon.com/dp/B017WJ5PR4` to `https://www.amazon.com/hz/audible/mlp/mfpdp/B017WJ5PR4` but that now redirects to `https://www.amazon.com/dp/B017WJ5PR4?ref=adbl_mfpdp_geo` again, at least for me. But if the cover art is limited to 500x500, I'll just remove the special case then.

kellnerd:
> I think we used to rewrite https://www.amazon.com/dp/B017WJ5PR4 to https://www.amazon.com/hz/audible/mlp/mfpdp/B017WJ5PR4 but that now redirects to https://www.amazon.com/dp/B017WJ5PR4?ref=adbl_mfpdp_geo again, at least for me.

Oops, then I was imply teasting with the wrong domain yesterday. This link also redirects for me.

Interestingly, the [larger image from our Audible test case](https://m.media-amazon.com/images/I/91WJ0q27ddL.jpg) still exists, but there seems to be no way to extract the image identifier from anywhere...

ROpdebee:
Hmm. Hopefully we find a way to extract those again later, but for now, it's better to have a 500x500 image than to have no image at all :)

ROpdebee:
> Interestingly, the [larger image from our Audible test case](https://m.media-amazon.com/images/I/91WJ0q27ddL.jpg) still exists, but there seems to be no way to extract the image identifier from anywhere...

Funnily enough, in #786, I changed the image extraction logic to use that of the physical product pages, and the old images are now in use again 😅

PR #721, opened 2023-12-08
fix: increase SPA ID length limit for work code validation

julian45:
SPA IDs are sequential numbers. Sometime within the last 2.5 years, the 10 millionth SPA ID was issued, which means that an increasing amount of IDs are now 8 characters long instead of 7. However, the script associated with this library currently only looks for 7 digits, and presents (incorrect) validation errors for 8-digit SPA IDs. This change aims to ameliorate that by increasing the character length the regex allows for.

### DIFF
--- a/lib/work_identifiers.js
+++ b/lib/work_identifiers.js
@@ -2,7 +2,7 @@
  * API for work identifiers. See inline docs.
  */
 const MBWorkIdentifiers = (() => {
-    const VERSION = '2021.5.27';  // Just for use in edit notes, really.
+    const VERSION = '2023.12.07';  // Just for use in edit notes, really.
 
     /**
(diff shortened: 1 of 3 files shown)


ROpdebee:
Thanks! Could you also update the `mb_bulk_copy_work_codes.user.js` and `mb_validate_work_codes.user.js` so that they use the newer version of the library? It should suffice to update the version number of the library in the `@require` (line 18 in both scripts), and also the `@version` (line 3 in both cases).

julian45:
[`695e56c`](https://github.com/ROpdebee/mb-userscripts/pull/721/commits/695e56c2619e8fac3049c1aedfc04d73da53f3c2) should do the trick.

ROpdebee:
Thanks for the fix!

PR #763, opened 2024-05-01
fix(inline recs): Crash when no existing div.ars

jesus2099:
I noticed that this userscript relies on mb_INLINE-STUFF to run properly.

Example release with no AR, no AcoustID and no ISRC (quick! it is being removed!)
https://musicbrainz.org/release/1f3a402f-7265-42ea-803a-f3cf8c508e7c

Or any other releases with credits at bottom and no mb_INLINE-STUFF running
https://musicbrainz.org/release/b4e8f7f4-e35c-42eb-a043-9605e4784aff

### DIFF
--- a/mb_qol_inline_recording_tracks.user.js
+++ b/mb_qol_inline_recording_tracks.user.js
@@ -1,6 +1,6 @@
 // ==UserScript==
 // @name         MB: QoL: Inline all recording's tracks on releases
-// @version      2024.5.1
+// @version      2024.5.2
 // @description  Display all tracks and releases on which a recording appears from the release page.
 // @author       ROpdebee
 // @license      MIT; https://opensource.org/licenses/MIT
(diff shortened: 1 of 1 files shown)


jesus2099:
I don't understand why my PR fails a test.
Initially, my commit was called `fix: Crash when no existing div.ars` and it failed no tests.
Then I renamed it to `fix(inline recs): Crash when no existing div.ars` and now it fails the `codecov/project` test. 🤔

jesus2099:
Initially I forked repo to see if I could exclude current track from the inline track list.
I felt the current track should not be displayed twice…
Then I found this problem. :D

jesus2099:
BTW I just noticed that [that release](https://musicbrainz.org/release/8fc541fb-7601-4a02-b859-b1d9789ddde7) had tracks with both `span.mp` and `span.name-variation`.
Those tracks are skipped as only one level of `span` is handled.
I added another commit.

kellnerd:
> I don't understand why my PR fails a test.

No worries, it is the [latest commit in main](https://github.com/ROpdebee/mb-userscripts/commit/65622a2114fa668a1b6f9284dfdd0d604fcb592b) which is currently failing the code coverage test (because a few Discogs tests had to be disabled temporarily).

jesus2099:
Sorry my second commit rejoice a nice thing from my first commit. 
As I'm in the toilets on GitHub and cannot restart my PC now and for some days maybe, I could not amend second commit send had to create a lame third commit. 😭

ROpdebee:
Thanks for the fixes!

jesus2099:
@ROpdebee, you forgot the https://github.com/ROpdebee/mb-userscripts/labels/mb_qol_inline_recording_tracks label, that you can also add to #764.

PR #766, opened 2024-05-28
fix: fix SoundCloud capitalization

Thermospore:
Minor thing, but I noticed SoundCloud was capitalized incorrectly as Soundcloud in the UI, so I decided to try and fix it...

I have no idea what I'm doing; hopefully this doesn't break stuff

Cheers!


ROpdebee:
Thanks for the patch! The changes you suggested were all correct, I only amended the commits to conform to the [commit message style adopted in this repository](https://www.conventionalcommits.org/en/v1.0.0/) and to replicate the capitalisation changes to the test files.

PR #764, opened 2024-05-02
feat(inline recs): Order tracks by ascending release dates

jesus2099:
For the moment, release tracks were displayed in random order.
Now they will be ordered by date, then title, then comment.

I have also added display of release comment and added release date as tooltip.

---

@ROpdebee, may I leave you manage the version number, OK?
My 3 concurrent PR have no conflict between themselves:

- This
- #776 
- #777

PR #776, opened 2024-07-20
fix(inline recs): Handle video recordings

jesus2099:
Remove `:first-child`, not needed and blocking video recordings detection

The selector is already strict enough to not catch recordings from `div.ars` (I guess it was the reason for those)

Examples where video data track was not handled before this commit: https://musicbrainz.org/release/d9e494dd-9386-48e4-8c52-9764dd8b8a0f
https://musicbrainz.org/release/0a5be94f-3429-46bd-9b1d-07dcb0209098

Release with recordings in inline relationships (on CD 23): https://musicbrainz.org/release/70dcde39-81ed-45e8-bdb3-f511112a3880

There was not problem with non-video data tracks:
https://musicbrainz.org/release/b63ca2ff-782e-45fc-bfb9-b74dc6aefe00

---

@ROpdebee, may I leave you manage the version number, OK?
My 3 concurrent PR have no conflict between themselves:

- #764
- This
- #777

### DIFF
--- a/mb_qol_inline_recording_tracks.user.js
+++ b/mb_qol_inline_recording_tracks.user.js
@@ -94,7 +94,7 @@ function insertRows(recordingTd, recordingInfo) {
 }
 
 function loadAndInsert() {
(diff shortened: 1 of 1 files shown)


kellnerd (approved):
Works for me, I've found no scenario where the previous overzealous selector were required.

PR #777, opened 2024-07-20
fix(inline recs): Fix mb.org exclude rules and add mb.eu match rule

jesus2099:
Exclude rule syntax is like include, not like match (//*.mb.org does not match //mb.org, only test.mb.org, beta.mb.org) Before this commit, exclude rules are not applied to main //mb.org website

I am also excluding more unrelated pages (open edits, edit history page 2+, details, cover arts, ...)

Add mb.eu support

---

@ROpdebee, may I leave you manage the version number, OK?
My 3 concurrent PR have no conflict between themselves:

- #764
- #776 
- This


kellnerd (approved):
Thank you, looks good to me.

I think I had also noticed the errors caused by this userscript running on release editor pages when I was debugging release seeding but was too lazy to fix this myself 😅

jesus2099:
Thanks for spotting the typo, I have now just fixed it in the commit text itself, as well.

PR #779, opened 2024-07-21
feat(caa dims): add support for new event art archive (EAA)

jesus2099:
@vzell request:

> https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/182
>
> Could the userscript **MB: Display CAA image dimensions** be extended to the new event-art-archive?

Tested on https://musicbrainz.org/event/d1bd6d40-55f6-4e1a-91df-4c9395f95401 and sub-pages (/event-art, /edits)

### DIFF
--- a/src/mb_caa_dimensions/image.ts
+++ b/src/mb_caa_dimensions/image.ts
@@ -83,7 +83,7 @@ export abstract class BaseImage {
 }
 
 function caaUrlToDirectUrl(urlObject: URL): URL {
-    if (urlObject.host.match(AA_DOMAINS) && urlObject.pathname.match(/^\/(event|release)\//)) {
+    if (AA_DOMAINS.test(urlObject.host) && /^\/(event|release)\//.test(urlObject.pathname)) {
         const [releaseId, imageName] = urlObject.pathname.split('/').slice(2);
         urlObject.href = `https://archive.org/download/mbid-${releaseId}/mbid-${releaseId}-${imageName}`;
     }
@@ -110,7 +110,7 @@ function urlToCacheKey(fullSizeUrl: string, thumbnailUrl?: string): string {
     // Ideally, the cache key for RG covers would be the full size URL of the release cover,
     // but we unfortunately cannot get the original image's extension here, so we cannot construct
     // it.
-    if (urlObject.host.match(AA_DOMAINS) && urlObject.pathname.startsWith('/release-group/')) {
+    if (AA_DOMAINS.test(urlObject.host) && urlObject.pathname.startsWith('/release-group/')) {
         assertDefined(thumbnailUrl, 'Release group image requires a thumbnail URL');
         return thumbnailUrl;
     }
@@ -156,7 +156,7 @@ function urlToDirectImageUrl(url: string): string {
 function parseCAAIDs(url: string): [string, string] {
     const urlObject = new URL(url);
 
-    if (urlObject.host.match(AA_DOMAINS) && urlObject.pathname.match(/^\/(event|release)\//)) {
+    if (AA_DOMAINS.test(urlObject.host) && /^\/(event|release)\//.test(urlObject.pathname)) {
         const [releaseId, thumbName] = urlObject.pathname.split('/').slice(2);
         const imageId = thumbName.match(/^(\d+)/)?.[0];
         assertDefined(imageId, 'Malformed URL');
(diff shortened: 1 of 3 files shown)


kellnerd:
/deploy-preview

You can use `npm run build` to build the userscripts locally.

I will see if I find some time to review your PRs as ROpdebee seems to be busy.

src/mb_caa_dimensions/image.ts
```diff
@@ -110,7 +110,7 @@ function urlToCacheKey(fullSizeUrl: string, thumbnailUrl?: string): string {
     // Ideally, the cache key for RG covers would be the full size URL of the release cover,
     // but we unfortunately cannot get the original image's extension here, so we cannot construct
     // it.
-    if (urlObject.host === CAA_DOMAIN && urlObject.pathname.startsWith('/release-group/')) {
+    if (AA_DOMAINS.test(urlObject.host) && urlObject.pathname.startsWith('/release-group/')) {
```
kellnerd, line 113: Since this code branch only deals with release group cover art, you can revert this change (and restore the CAA_DOMAIN constant in addition to the regex).

kellnerd (commented):
Thank you, I've tested the preview and everything works as expected.
Regarding the code I have one minor suggestion.

P.S. There are a few more places where the wording could be changed from specifically CAA to the more general "image archives" (the name which is also used [on MetaBrainz Jira](https://tickets.metabrainz.org/browse/IMG)), but that is not required here and I will leave that decision up to ROpdebee.

src/mb_caa_dimensions/image.ts
jesus2099: Why not.
But that would make two constants (`CAA_DOMAIN` and `AA_DOMAINS`\*) instead of one.
What do you think @ROpdebee?

\* `AA_DOMAINS`should probably become `IMG_DOMAINS`, see https://github.com/ROpdebee/mb-userscripts/pull/779#pullrequestreview-2190285233
`CAA` project was renamed to `IMG` in https://tickets.metabrainz.org where `IMG = CAA + EAA` 😉

jesus2099: I made this change and also renamed my made-up AA into IMG, like in https://tickets.metabrainz.org/projects/IMG to answer [your P.S.](https://github.com/ROpdebee/mb-userscripts/pull/779#pullrequestreview-2190285233).
Please check.

I didn't manage tu build (`npm run build` failed). :)
There is a kind of install step before, like `npm install --save-dev` or something like that?

jesus2099:
> You can use `npm run build` to build the userscripts locally.

I didn't manage tu build (`npm run build` failed). :-/

```sh
$ npm run build

> mb-userscripts@1.0.0 build
> NODE_ENV=production tsx build/build.ts

'NODE_ENV' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
```

I tried some tips from stackoverflow but nevermind.

jesus2099:
> P.S. There are a few more places where the wording could be changed from specifically CAA to the more general "image archives" (the name which is also used [on MetaBrainz Jira](https://tickets.metabrainz.org/browse/IMG)), but that is not required here and I will leave that decision up to ROpdebee.

I renamed my made-up AA into IMG, like in https://tickets.metabrainz.org/projects/IMG 
Please check.

kellnerd:
/deploy-preview

Looks good to me. I will give @ROpdebee a chance to react before I will merge, though.

> I didn't manage tu build (npm run build failed). :-/

Your terminal does not support the `NODE_ENV=production` syntax before the actual command (which just sets an environment variable), this happens on Windows computers for example.

Alternatively you can also run `npm run build-dev` which is basically the same command but without the environment variable. The difference doesn't matter for local testing, it just skips minification of some code blocks IIRC.
(And yes, you might have to run `npm install` if your `node_modules/` folder is not up-to-date.)

kellnerd (approved):
Let's not delay this feature further, it is definitely good enough for the initial release.
Code clean-up can be done later (or not).

jesus2099:
Which code cleanup?

kellnerd:
Oh, maybe cleanup wasn't the best term, what I meant is rather renaming/rewording.
Like updating all identifiers and comments which are still talking about CAA and releases but also cover EAA and events now.

jesus2099:
Ah OK! I forgot about that. :)

PR #780, opened 2024-07-23
build: Fix dynamic import of metadata modules on Windows (again)

kellnerd:
I already did this in #34, but it was reverted when we switched from ESM to CJS in #147.
Now that a2e7fc7 reintroduced ESM with `tsx`, we need this patch again.

See the problem discussion at https://github.com/ROpdebee/mb-userscripts/pull/779#issuecomment-2242718714 and https://github.com/kellnerd/userscript-bundler/issues/4.


kellnerd:
A quick check under Linux confirmed that the build process is still working as well.

Issue #804, opened 2024-12-19
ACUM ID format


dvirtz:
The batch copy work codes script should remove leading zeros from ACUM work ID. 
For example, https://musicbrainz.org/work/bd895f6a-4f4d-4d0f-8f91-b93b847def43 ACUM ID was entered as 00001214292001
while the read ID is 1214292001: https://nocs.acum.org.il/acumsitesearchdb/version?workid=1214292&versionid=1214292001

PR #805, opened 2024-12-19
fix(work codes): add ACUM ID

dvirtz:
fixes ACUM ID format #804

### DIFF
--- a/lib/work_identifiers.js
+++ b/lib/work_identifiers.js
@@ -2,7 +2,7 @@
  * API for work identifiers. See inline docs.
  */
 const MBWorkIdentifiers = (() => {
-    const VERSION = '2023.12.07';  // Just for use in edit notes, really.
+    const VERSION = '2025.03.09';  // Just for use in edit notes, really.
 
     /**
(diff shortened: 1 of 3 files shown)


kellnerd (approved):
I will trust your expertise here.
Let me know if the version bump from the other PR is sufficient to also update this `@require`'d file.

kellnerd:
Hmm, I can't merge this because some checks are not running... As we have to wait/resolve this anyway, you could push a commit which updates the version in the require URLs of the two scripts which are using this library.

dvirtz:
Actually, since opening this PR I learned that ACUM can also have letters in their identifiers as shown by https://github.com/dvirtz/musicbrainz-scripts/issues/56.
I updated the regex and versions.

kellnerd:
Thank you, with your last push the button to approve CI workflows finally reappeared.
Can you also bump the versions of the two userscripts themselves, please, otherwise nobody will get notified about the updated library. You probably have to rebase on `main` to avoid a merge conflict because of yesterday's release.

dvirtz:
Done

PR #807, opened 2025-02-14
fix(Bulk copy-paste work codes) issues on https://iswcnet.cisac.org/

Lioncat6:
Replace field `PREFERRED_ISWC_FIELD` (Which seemingly doesn't exist anymore) with `ISWC_FIELD`

image


Lioncat6:
Need to re-name commit

PR #808, opened 2025-02-14
fix(work codes): Update ISWC table selector on https://iswcnet.cisac.org/search

Lioncat6:
Replaced table field selector `PREFERRED_ISWC_FIELD` with `ISWC_FIELD`

image


dvirtz:
@kellnerd is this something you can merge?

kellnerd (approved):
Yes, I can.
As I am not using this script myself, I am approving the PR under the assumption that CISAC no longer provides alternative ("archived") ISWC values.
Seems to work for me now, thank you @Lioncat6!

PR #810, opened 2025-03-04
feat(ecau): Seed cover art from Harmony release actions pages

YoGo9:
Added support for harmony.pulsewidth.org.uk to the Enhanced Cover Art Uploads userscript. This allows users to easily add cover art from the Harmony importer to MusicBrainz releases with a single click, similar to the existing a-tisket integration.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/meta.ts
+++ b/src/mb_enhanced_cover_art_uploads/meta.ts
@@ -18,6 +18,7 @@ const metadata: UserscriptMetadata = {
         '*://atisket.pulsewidth.org.uk/*',
         '*://etc.marlonob.info/atisket/*',
         '*://vgmdb.net/album/*',
+        '*://harmony.pulsewidth.org.uk/*',
     ],
     'exclude': ['*://atisket.pulsewidth.org.uk/'],
     'grant': [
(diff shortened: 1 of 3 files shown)


YoGo9:
I've compiled a version with these changes that you can test here:

https://github.com/YoGo9/mb-userscripts/raw/refs/heads/working/dist/mb_enhanced_cover_art_uploads.user.js

kellnerd (approved):
Thank you!

I've fixed a few things which were causing CI checks to fail:
- Reworded your commit messages to follow https://www.conventionalcommits.org/en/v1.0.0/ and squashed your first three commits.
- Stylistic changes to make eslint happy.

Additionally I've turned the clickable `div` containers into proper HTML links, which the users can open in the same or a new tab, as they prefer. This also gives them a blue background and the appropriate cursor, so the injected CSS is no longer necessary.
And finally I'm manipulating the origin URL to point to a cached version of the Harmony release actions page (leveraging an undocumented feature 😏).

Two potential improvements remain, but they can be done in subsequent PRs:
- Display image dimensions. We can probably reuse the entire `atisket/dimensions.ts` module for that.
- Seed the release URL instead of the image URL. This would allow ECAU to grab multiple images for supported providers (currently: Bandcamp) and we can remove the assumption that the image is a front cover (which is not always true, Bandcamp again).

YoGo9:
Amazing!

PR #816, opened 2025-05-06
fix(ecau): upstream provider fixes

ROpdebee:
- Remove DatPiff as it's been down for at least a year now, with no sign when it'll return.
- Add support for the Panel cover art type (only used to copy existing types when cloning artwork from an existing MB release).
- Use larger thumbnails when deduplicating Bandcamp images. Somehow the 150x150 images were different even though the source images were the same. This isn't the most reliable fix though, see commit message for possible future fixes.
- Change atisket image file size tests to align with changed file sizes upstream.
- Updated browserslist as it kept complaining.

Not fixed/not checked:
- VGMdb.info API tests, as the API is down (as it usually is; maybe we should remove it).
- One of the soundcloud test releases has been removed, haven't find a replacement (maybe the test should be disabled?).
- VK Music now requires an API, so the provider is effectively broken. As nobody has complained about it, I'm not going to prioritise fixing it.
- Yandex Music is unavailable in my region, so couldn't load up-to-date pages.


ROpdebee:
> Hi, cool to see you back working on MB scripts!

The comeback might be shortlived though 😅 

> While you are already at it, could you also update the bundled Image Max URL version? There seems to be a [new trick for AllMusic](https://community.metabrainz.org/t/allmusic-and-covers-trick/756685/26?u=kellnerd) cover images.

Done! Also updated the favicon, seems like it was broken.

PR #841, opened 2025-05-14
build: update minimum browser versions

ROpdebee:
We're using some ES2022 features, so let's make sure we compile to browsers that support those features, thus maybe transpiling less. Let's also configure TypeScript to use typings limited to those browsers, otherwise it might allow us to use things like `Iterator#map`, which are barely available in current browsers. That'll also avoid the `Array#at` situation from a 
few years back.

Speaking of `Array#at`, seeing as all major browsers have supported it for several years now, let's start embracing it too.


ROpdebee:
`nativejsx` is throwing another spanner in the works. It depends on `esbuild` to output the transformed code back into a string, but `esbuild` hasn't been maintained in a while and doesn't support newer JS syntax, which is causing compilation to fail.

I think it's time to switch away from `nativejsx`. That needs to happen before this PR can pass.

PR #849, opened 2025-09-24
fix(caa_dim): Use multiplication sign instead of letter x

jesus2099:
Then 640×480 looks more legit than 640x480 pixels


ROpdebee:
Thanks!

PR #850, opened 2025-09-30
fix(spotify): remove User-Agent header

chaban-mb:
As [described in the forum](https://community.metabrainz.org/t/ropdebees-userscripts-support-thread/551947/210?u=chaban), Spotify is messing with the Open Graph tags. Apparently simply omitting the User-Agent header fixes this problem.

### DIFF
--- a/src/mb_enhanced_cover_art_uploads/providers/spotify.ts
+++ b/src/mb_enhanced_cover_art_uploads/providers/spotify.ts
@@ -1,6 +1,7 @@
 import { qsMaybe } from '@lib/util/dom';
 
 import { HeadMetaPropertyProvider } from './base';
+import type { RequestOptions } from '@lib/util/request';
 
 export class SpotifyProvider extends HeadMetaPropertyProvider {
     public readonly supportedDomains = ['open.spotify.com'];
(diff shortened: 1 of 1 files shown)


src/mb_enhanced_cover_art_uploads/providers/spotify.ts
```diff
@@ -1,6 +1,7 @@
 import { qsMaybe } from '@lib/util/dom';
 
 import { HeadMetaPropertyProvider } from './base';
+import type { RequestOptions } from '@lib/util/request';
```
ROpdebee, line 4: This needs to be moved to the very first line in the file.

src/mb_enhanced_cover_art_uploads/providers/spotify.ts
```diff
@@ -11,4 +12,14 @@ export class SpotifyProvider extends HeadMetaPropertyProvider {
     protected override is404Page(document_: Document): boolean {
         return qsMaybe('head > meta[property="og:title"]', document_) === null;
     }
+
+    protected override fetchPage(url: URL, options?: RequestOptions): Promise<string> {
+        return super.fetchPage(url, {
+            ...options,
+            headers: {
+                ...options?.headers,
+                'User-Agent': null,
```
ROpdebee, line 21: Does it also work when you provide an empty string? Otherwise we need to suppress the type error raised by TypeScript somehow.

src/mb_enhanced_cover_art_uploads/providers/spotify.ts
```diff
@@ -11,4 +12,14 @@ export class SpotifyProvider extends HeadMetaPropertyProvider {
     protected override is404Page(document_: Document): boolean {
         return qsMaybe('head > meta[property="og:title"]', document_) === null;
     }
+
+    protected override fetchPage(url: URL, options?: RequestOptions): Promise<string> {
+        return super.fetchPage(url, {
```
ROpdebee, line 17: Could you add a code comment linking to the forum post here? Otherwise I'll probably forget why this override was needed.

